PR #44, opened 2025-09-01
MRB-471: Inference with interpolator model

dnerini:


OpheliaMiralles (approved):
Looks OK to me! + the command creates grib files as expected

PR #47, opened 2025-09-15
Support for time interpolation on analyses

OpheliaMiralles:
## Objective
Enable time interpolation runs from the analysis instead of from a forecaster run. This enables further insights into performance of the time interpolator.

## Changes
- New configs
  - interpolation from analysis
  - forecaster without trim edge (needed for COSMO-1E)
- Separation of run_id and mlflow_id to enable multiple runs with the same model for e.g.
  - interpolator run from analysis and forecaster
  - interpolator run with different forecasters
- Automatic run_ids constructed from first 8 characters of mlflow_id e.g.
  - `9c18b9007-analysis` for the interpolator (M-2) run from analysis 
  - `9c18b9007-d0846032f` for the interpolator (M-2) run from a forecaster (M-1)

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -2,11 +2,7 @@ import copy
 from datetime import datetime, timedelta
 import yaml
 import hashlib
-import logging
 import json
-
-LOG = logging.getLogger(__name__)
-logging.basicConfig(level=logging.WARNING)
 
 CONFIG_ROOT = Path("config").resolve()
 OUT_ROOT = Path(config["locations"]["output_root"])
@@ -91,8 +87,6 @@ def collect_all_runs():
 
         # Register this (possibly composite) run inside the loop
         runs[run_id] = run_config
-        LOG.warning(f"Registered run '{run_id}' (model_type={run_config['model_type']})")
-
     return runs
 
 


workflow/scripts/set_inference_pyproject.py
```diff
@@ -374,6 +373,36 @@ def update_pyproject_toml(
...
+            if "mlflow_id" in conf:
+                all_ids[conf["run_id"]] = conf["mlflow_id"]
+            # if the forecaster key exists (interpolator usecase) and is not None (interpolator from analysis), register it too otherwise pyproject does not get created
+            if (
+                "forecaster" in conf
+                and conf["forecaster"] is not None
+                and "mlflow_id" in conf["forecaster"]
+            ):
+                all_ids[conf["forecaster"]["run_id"]] = conf["forecaster"]["mlflow_id"]
+    logger.info("All run IDs with MLflow IDs: %s", all_ids)
```
jonasbhend, line 402: This bugs me that we seem to have to do it over and over again. I think we should be able to do it once (at start of snakemake run) in commons.smk and then re-use the RUN_CONFIGS object. Having said that, it didn't work when I tried, but this may also have been due to other issues.

jonasbhend: not necessarily something we need to fix right now. We already do config parsing multiple times in commons.smk, that could probably be streamlined a lot too.

workflow/rules/common.smk
```diff
@@ -74,9 +78,15 @@ def collect_all_runs():
...
         run_id = run_config.pop("run_id")
         runs[run_id] = run_config
+        runs[run_id] = run_config
         if model_type == "interpolator":
-            run_id = run_config["forecaster"]["run_id"]
-            runs[run_id] = run_config["forecaster"]
+            if "forecaster" not in run_config or run_config["forecaster"] is None:
+                logging.warning(
+                    f"Interpolator '{run_id}' has not been set a 'forecaster' config. Please use the 'interpolator_from_files.yaml' config to run an interpolator without a forecaster."
+                )
```
jonasbhend, line 86: I think this warning is superfluous. We should not warn if the interpolator is properly specified, but raise an error if not.

If the interpolator_from_files.yaml is not specified for forecasting from analysis, this will trigger an error at inference time. If we want to provide a more helpful log entry, fine, but I think otherwise we could also completely ditch this.

jonasbhend (commented):
Hi Ophelia, this is looking good. Thanks a lot for your help. I have only minor comments. Could you please address these?

workflow/scripts/set_inference_pyproject.py
OpheliaMiralles: Yes we should find a more clever way to do it for sure. Maybe we'll open a branch feature/optimize_config_parsing where we dive into the issue, I guess otherwise the scope of this PR will be a bit large

README.md
```diff
@@ -36,16 +36,13 @@ dates:
...
 
 runs:
-  # - interpolator:
-  #     run_id: 9c18b90074214d769b8b383722fc5a06
-  #     label: Interpolator (M-1 forecaster)
-  #     forecaster:
-  #       run_id: d0846032fc7248a58b089cbe8fa4c511
   - forecaster:
-      run_id: 2f962c89ff644ca7940072fa9cd088ec
+      run_id: forecaster-stage-d
```
dnerini, line 40: could we compute this internally as a combination of the label and the mlflow run id? something like

```
run_id = f"{label.replace(' ', '_')}__{mlflow_id[:8]}"
```

OpheliaMiralles: I think @jonasbhend is on it right now

dnerini: sounds good, but I would also try to make it more human readable (by using the label for example), as currently it's really hard to navigate the output folders and log files with only hashes in their names ...

jonasbhend: No, I don't think using the labels is a good idea. These are basically just free text. If you want to change the free text, you trigger a complete rerun. That's why I was using something different. What we could be using is the mlflow description (if there is such a thing) and then paste forecaster and interpolator together. But definitely not the label as this is bound to change more often (probably for each experiment).

jonasbhend: I also fear that the solution with the automatic run_id is somewhat harder to implement, as we are currently parsing configs at multiple places (in scripts, etc.). So we need to be careful to do this only once (or with the same function each time).

Unfortunately I am not working today, so won't be able to make progress on this. If you want I can push my current state to a separate branch or this PR if we are only going to merge this with the automatic run_ids. Just let me know.

jonasbhend:
Currently the pipelines fail with:

```
(evalml) (base) bhendj@balfrin-ln004:/scratch/mch/bhendj/evalml> evalml experiment config/interpolators.yaml --report
Launching: snakemake --executor slurm --default-resources slurm_partition=postproc cpus_per_task=1 mem_mb_per_cpu=1800 runtime=1h --jobs 10 --configfile config/interpolators.yaml --cores 4 --report-after-run --report experiment_report.html experiment_all
INFO:snakemake.logging:None
INFO:snakemake.logging:host: balfrin-ln004
INFO:snakemake.logging:Building DAG of jobs...
host: balfrin-ln004
Building DAG of jobs...
ERROR:snakemake.logging:MissingInputException in rule report_experiment_dashboard in file "/scratch/mch/bhendj/evalml/workflow/rules/report.smk", line 10:
Missing input files for rule report_experiment_dashboard:
    output: output/results/74795720/metrics/dashboard
    wildcards: experiment=74795720
    affected files:
        output/data/runs/9c18b9007-analysis/verif.aggregated.nc
        output/data/runs/9c18b9007-d0846032f/verif.aggregated.nc
        output/data/baselines/COSMO-E-1h/verif.aggregated.nc
        output/data/runs/d0846032f/verif.aggregated.nc
        output/data/runs/07c3d9698-d0846032f/verif.aggregated.nc
MissingInputException in rule report_experiment_dashboard in file "/scratch/mch/bhendj/evalml/workflow/rules/report.smk", line 10:
Missing input files for rule report_experiment_dashboard:
    output: output/results/74795720/metrics/dashboard
    wildcards: experiment=74795720
    affected files:
        output/data/runs/9c18b9007-analysis/verif.aggregated.nc
        output/data/runs/9c18b9007-d0846032f/verif.aggregated.nc
        output/data/baselines/COSMO-E-1h/verif.aggregated.nc
        output/data/runs/d0846032f/verif.aggregated.nc
        output/data/runs/07c3d9698-d0846032f/verif.aggregated.nc
```

workflow/rules/common.smk
```diff
@@ -2,8 +2,12 @@ import copy
 from datetime import datetime, timedelta
 import yaml
 import hashlib
+import logging
 import json
 
+LOG = logging.getLogger(__name__)
+logging.basicConfig(level=logging.WARNING)
```
dnerini, line 9: this reconfigures the root logger and overrides Snakemake’s console logger, is it intended? this results in all those `INFO:snakemake.logging:...` lines in the terminal output

workflow/rules/inference.smk
```diff
@@ -185,14 +192,18 @@ rule inference_interpolator:
         reftime_to_iso=lambda wc: datetime.strptime(
             wc.init_time, "%Y%m%d%H%M"
         ).strftime("%Y-%m-%dT%H:%M"),
-        forecaster_run_id=lambda wc: _get_forecaster_run_id(wc.run_id),
+        forecaster_run_id=lambda wc: (
+            "null"
+            if RUN_CONFIGS[wc.run_id].get("forecaster") is None
+            else RUN_CONFIGS[wc.run_id]["forecaster"]["run_id"]
```
dnerini, line 198:
```suggestion
            else RUN_CONFIGS[wc.run_id.split("-")[1]]
```

workflow/scripts/verif_baseline.py
```diff
@@ -154,7 +154,7 @@ def main(args: ScriptConfig):
         )
     baseline = baseline[args.params].sel(
         ref_time=args.reftime,
-        lead_time=np.array(args.lead_time, dtype="timedelta64[h]"),
+        lead_time=np.array(args.lead_time, dtype="timedelta64[h]"), method="nearest"
```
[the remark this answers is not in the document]
jonasbhend: We can already deal with different lead times for different forecasts (e.g. forecaster and interpolator)

workflow/rules/inference.smk
jonasbhend: I don't think this works. But maybe the problem is indeed that we haven't generated the run_id for the forecaster in the interpolator config (mind you, as long as you don't pop the forecaster, the whole forecaster config is still present in RUN_CONFIGS, which is what i used here).

workflow/rules/common.smk
```diff
@@ -84,19 +86,22 @@ def collect_all_runs():
                     f"Interpolator '{run_id}' has no forecaster; using analysis inputs."
                 )
             else:
-                tail_id = run_config["forecaster"]["mlflow_id"][0:9]
+                tail_id = run_config["forecaster"]["mlflow_id"]
                 # Ensure a proper 'forecaster' entry exists with model_type
                 fore_cfg = copy.deepcopy(run_config["forecaster"])
                 fore_cfg["model_type"] = "forecaster"
-                runs[tail_id] = fore_cfg
+                runs[codenamize(tail_id)] = fore_cfg
```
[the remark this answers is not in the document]
jonasbhend: Reverted the human-readable run_ids. Could be implemented in a separate PR if this is really what we want.

config/interpolators.yaml
```diff
@@ -3,36 +3,44 @@ description: |
...
 lead_time: 120h
 
 runs:
   - interpolator:
-      run_id: 9c18b90074214d769b8b383722fc5a06
+        mlflow_id: 9c18b90074214d769b8b383722fc5a06
+        label: LAM Interpolator (COSMO-E analysis)
+        steps: 0/121/1
+        config: resources/inference/configs/interpolator_from_files.yaml
+        forecaster: null
```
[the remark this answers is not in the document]
OpheliaMiralles: Yes but I agree that it would actually be nice to be able to interpolate e.g. COSMO-2E. But it would require an additional pipelines where the data is taken from grib COSMO-2E. I don't think it's impossible (I never do) but maybe out of scope for this PR

resources/inference/configs/interpolator_from_files.yaml
jonasbhend, line 1: Can we rename this? As far as I understand we always interpolate from files, so maybe we can find a better name. My proposal would be to rename to `interpolator_from_analysis.yaml` or if this is indeed what we are doing `interpolator_from_test_data.yaml`

config/interpolators.yaml
```diff
@@ -4,35 +4,43 @@ description: |
...
       steps: 0/121/1
       config: resources/inference/configs/interpolator_stretched.yaml
       forecaster:
-        run_id: d0846032fc7248a58b089cbe8fa4c511
+        mlflow_id: d0846032fc7248a58b089cbe8fa4c511
         config: resources/inference/configs/forecaster_with_global.yaml
       extra_dependencies:
-         - git+https://github.com/MeteoSwiss/anemoi-inference@fix/interp_files
-        #  - git+https://github.com/ecmwf/anemoi-inference@fix/interp_files
+        - git+https://github.com/MeteoSwiss/anemoi-inference@fix/interp_files
```
OpheliaMiralles, line 43:
```suggestion
         - git+https://github.com/MeteoSwiss/anemoi-inference@fix/interp_files
```

workflow/rules/common.smk
```diff
@@ -2,8 +2,12 @@ import copy
 from datetime import datetime, timedelta
 import yaml
 import hashlib
+import logging
```
jonasbhend, line 5: I am still not sure we need logging here. The warnings below should at most be debug statements imho and indeed probably can be inferred from the snakemake logs anyway, no?

frazane (approved):

PR #48, opened 2025-09-15
Independent run_ids

jonasbhend:
This is a WIP to test the concept of using independent run_id and mlflow_ids. 

## Objective
Separate the run_id used to identify a snakemake forecast run used in an experiment from the mlflow id allows us to run different combinations with the same model. For example one could run the interpolator with mlflow_id=aaabb... once from the forecaster and once from analysis. Or run the same interpolator with different forecasters etc.


jonasbhend:
Ok, @dnerini @OpheliaMiralles this now somewhat works, but I got stuck at trying to run the forecaster with configuration `null`. Apparently somewhere in the DAG we should make sure that snakemake understands not to try to run inference with the forecaster when we run the interpolator from analysis. Any pointers?

PR #49, opened 2025-09-17
Add setting for global resources

dnerini:
Currently, we can indicate the number of parallel SLURM jobs via the config file. However, in principle this number might change depending on the resource being used. For example, we should control the maxim number of GPUs used in parallel, as we have to follow internal guidelines. 

This PR introduces global resources that apply across all submissions. The relative snakemake documentation for this feature is available here https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources-and-remote-execution

>  resources may represent either a global constraint across all submissions (e.g. number of API calls per second), or a constraint local to each specific job sumbmission (e.g. the amount of memory available on a node). [...] By default, mem_mb, disk_mb, and threads are all considered "local" resources, meaning specific to individual submissions. [...] All other resources are considered "global", meaning they are tracked across all jobs across all submissions. For example, if api_calls was limited to 5 and each job scheduled used 1 api call, only 5 jobs would be scheduled at a time, even if more job submissions were available.

In our case, we define a new `gpus=15` (global) resource and set `gpus=1` for the inference rules, so that snakemake will schedule at max 15 concurrent jobs on gpu nodes.

### DIFF
--- a/workflow/rules/inference.smk
+++ b/workflow/rules/inference.smk
@@ -133,6 +133,7 @@ rule inference_forecaster:
         runtime="20m",
         gres="gpu:1",
         slurm_extra=lambda wc, input: f"--uenv={Path(input.image).resolve()}:/user-environment",
+        gpus=1,
     shell:
         """
         (
@@ -206,6 +207,7 @@ rule inference_interpolator:
         runtime="30m",
         gres="gpu:1",
         slurm_extra=lambda wc, input: f"--uenv={Path(input.image).resolve()}:/user-environment",
+        gpus=1,
     shell:
         """
         (
--- a/config/forecasters.yaml
+++ b/config/forecasters.yaml
@@ -37,9 +37,12 @@ locations:
 
 profile:
   executor: slurm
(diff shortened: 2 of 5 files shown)


workflow/rules/inference.smk
```diff
@@ -132,6 +132,7 @@ rule inference_forecaster:
         runtime="20m",
         gres="gpu:1",
         slurm_extra=lambda wc, input: f"--uenv={Path(input.image).resolve()}:/user-environment",
+        gpus=1,
```
OpheliaMiralles, line 135: Would it be possible to also add the flexibility to chose the number of GPUs for each model @dnerini or is it out of scope for this PR?

dnerini: I'd say out of scope since this is really about limit the total number of GPUs, but what you suggest can be easily implemented in a separate PR

dnerini: see https://github.com/MeteoSwiss/evalml/pull/53

PR #51, opened 2025-09-25
Add GRIB templates for COSMO-1E 

frazane:


dnerini:
GitHub actions are still disabled for the whole org following the NPM supply chain attack (see infonet article)

PR #52, opened 2025-09-26
Evaluation workflow for 1km model

OpheliaMiralles:
Draft branch with working inference for COSMO-1. An example config is included in config. Working with a hack: template references are up for COSMO-1E but ideally, the template should be assigned with regards to the number of points on the grid (or other criterion).

### DIFF
--- a/workflow/scripts/verif_aggregation.py
+++ b/workflow/scripts/verif_aggregation.py
@@ -8,7 +8,9 @@ import numpy as np
 import xarray as xr
 
 LOG = logging.getLogger(__name__)
-logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+)
 
 
 def get_season(time):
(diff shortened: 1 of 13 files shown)


OpheliaMiralles:
Thanks @frazane for progressing on that! Is it in a working state or is there some more work to do?

workflow/scripts/verif_from_grib.py
```diff
@@ -184,7 +196,7 @@ def main(args: ScriptConfig):
         (datetime.now() - start).total_seconds(),
         fct,
     )
-
+    print("FCT TIMES", fct.time.values, flush=True)
```
OpheliaMiralles, line 199: Debugging residual

workflow/tools/config.schema.json
```diff
@@ -313,6 +313,19 @@
           "default": null,
           "description": "Number of GPUs to request.",
           "title": "Gpu"
```
OpheliaMiralles, line 315: Here, why is "Gpu" with only capital G? In description it is full caps and in the config all lower case, maybe we should chose one or the other

workflow/scripts/verif_aggregation.py
```diff
@@ -112,3 +114,4 @@ def main(args: Namespace) -> None:
     main(args)
     # example usage:
     # uv run workflow/scripts/verif_results.py /users/fzanetta/projects/mch-anemoi-evaluation/output/7c58e59d24e949c9ade3df635bbd37e2/*/verif.csv
+    # uv run workflow/scripts/verif_results.py /users/fzanetta/projects/mch-anemoi-evaluation/output/7c58e59d24e949c9ade3df635bbd37e2/*/verif.csv
```
OpheliaMiralles, line 117:
```suggestion
```

PR #54, opened 2025-10-03
Refactor plot rule and add interactivity

dnerini:
I initially set out to try marimo for some interactive work on the plotting PR. But while working on it, I realized that the script was doing quite a lot: plotting maps for all parameters, lead times, and regions, then stitching everything into an animation. On top of that, it included plenty of boilerplate (creating folders, running subprocesses, …) which are essentially tasks that Snakemake is designed to handle for us.

So instead of just adding interactivity, I refactored the code to lean on Snakemake’s strengths. The result is hopefully a cleaner workflow that showcases how Snakemake can simplify orchestration, remove boilerplate, and make the pipeline easier to maintain and extend.

Thoughts?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "netcdf4>=1.7.2",
     "cartopy",
     "earthkit-plots",
+    "marimo>=0.16.5",
 ]
 
 [project.optional-dependencies]
--- a/uv.lock
+++ b/uv.lock
@@ -1,9 +1,23 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.12'",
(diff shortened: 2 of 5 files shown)


workflow/notebooks/plot_forecast_frame.py
```diff
@@ -0,0 +1,150 @@
...
+        nrows=1, ncols=1, projection=projection, region=region, size=(8,8),
+    )
+    subplot = fig.add_map(row=0, column=0)
+
+    if param == "uv":
+        field = np.sqrt(
+            state["fields"]["10u"] ** 2 + state["fields"]["10v"] ** 2
+        )
+    elif param == "2t":
+        field = state["fields"][param] - 273.15
```
OpheliaMiralles, line 129: could add 2d too I guess if the objective is to have temperature in degrees C

PR #62, opened 2025-10-15
Adopt 'steps' instead of 'lead_time'

dnerini:


frazane (commented):
Don't we still somehow need a top-level specification of the step/leadtime? Otherwise, where do we get the information on what data to consider for verification?

jonasbhend:
Thanks a lot @dnerini for cleaning this up. This PR is exactly what I had in mind when working on the infamous #43. I just didn't want to touch that part as well back then, but I think it makes a lot of sense to specify this on a per-run basis for maximum flexibility.

config/forecasters-co1e.yaml
jonasbhend: Here the forecast range of the baseline is only +33h. Maybe this doesn't create an issue in the evaluation run as we only use what is there, but I guess it would be more prudent and transparent to specify the steps for COSMO-1E baseline as `0/36/6`

src/evalml/config.py
```diff
@@ -86,6 +90,29 @@ class RunConfig(BaseModel):
...
+            raise ValueError("Steps must be formatted as 'start/end/step'.")
+        try:
+            start, end, step = map(int, parts)
+        except ValueError:
+            raise ValueError("Start, end, and interval must be integers.")
+        if start >= end:
+            raise ValueError(f"Start ({start}) must be less than end ({end}).")
+        if step <= 0:
+            raise ValueError(f"Interval ({step}) must be a positive integer.")
+        if (end - start) % step != 0:
```
jonasbhend: The help above says that the definition is `[start, end)`, so this should be equivalent `0/126/6` to `0/121/6`, but it is not. We either need to adjust the documentation or change the paradigm. I am (obviously) much in favor for intervals including the end point (if matching with steps) such that we can just specify `0/120/6` for COSMO-(2)E and `0/33/6` for COSMO-1(E) without thinking to hard ;-).

jonasbhend (approved):
only minor comments, I approve once these are addressed

src/evalml/config.py
dnerini: agreed!

PR #63, opened 2025-10-15
Update configs

OpheliaMiralles:
Changes during review of anemoi-inference PR involved to remove the key `forcings` from the config but replace it with `constant_forcings`, which was what was needed in our case but avoid unnecessary changes in anemoi.
Also added new config to run stretched interpolator from test data.


resources/inference/configs/interpolator_from_test_data_stretched.yaml
dnerini: what is this new config about? is it used?

dnerini (approved):

PR #64, opened 2025-10-15
Distinguish between primary runs ('candidates') and secondary runs

dnerini:
Currently results are produced for ALL runs specified in the evalml config, including forecasters that are there only to produce inputs for interpolators. Instead, only "primary" runs should be listed. I call now those "candidates".


frazane (approved):

jonasbhend (approved):
Looking good.

PR #69, opened 2025-10-20
Update vega-lite spec for verification dashboard

frazane:
- fixes the tick labels of the x (lead time) axis which were overlapping
- adds interactivity: one can now zoom and move the x axis to focus on specific ranges of the lead time, and the range of the y axis will adapt to the data in each plot automatically
- (extra) formatting

### DIFF
--- a/resources/report/dashboard/script.js
+++ b/resources/report/dashboard/script.js
@@ -1,11 +1,11 @@
 // Tab switching
 document.querySelectorAll(".tab-link").forEach(button => {
-    button.addEventListener("click", () => {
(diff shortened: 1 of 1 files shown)


frazane:
> @frazane can we reduce this to be just about the vega-lite spec? I think this is fairly minimal and should be able to merge instantly. Please let me know if you want me to have a go (I'd just cherry-pick [866434e](https://github.com/MeteoSwiss/evalml/commit/866434ebe5a4d335edddbd11f7082297ca4b91c8) on top of main).

@jonasbhend  this is branched off #68. Once that is merged, changes compared to main will be minimal. This is also why it's still marked as draft, I am waiting for the other PR.

PR #71, opened 2025-10-20
Adopt forecast intervals including the end point

dnerini:
See https://github.com/MeteoSwiss/evalml/pull/62#discussion_r2444505781

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ description: |
 dates:
   start: 2020-01-01T12:00
   end: 2020-01-10T00:00
-  frequency: 54h
+  frequency: 60h
 
 runs:
   - forecaster:


jonasbhend (approved):
good, (one suggestion for COSMO-E init times)

jonasbhend:
I am just running the example configs to make sure everything works.

workflow/scripts/extract_baseline.py
```diff
@@ -54,7 +54,7 @@ def extract(
         raise ValueError("Currently only COSMO-E and COSMO-1E are supported.")
     tar_archive = tarfile.open(tar, mode="r:*")
     out = ekd.SimpleFieldList()
-    for lt in lead_time:
+    for lt in lead_times:
         filename = f"{tar.stem}/grib/{gribname}{lt:03}_{run_id}"
```
jonasbhend: Maybe we should change the naming to steps here to, no?

PR #72, opened 2025-10-21
Mrb 550 inconcsistent forecast initializations in evalml

jonasbhend:
To avoid comparing apples with oranges, we need to make sure that the exact initialization specified in the config are used for evaluation for both the forecasts and baselines. This was previously not the case for the baselines which used the nearest available initialization in the dataset.

## Changes
* exact selection of ref_times in baseline
* fix configs for availability of COSMO-1E baseline (starting in 2020-08-01)

### DIFF
--- a/config/forecasters-co1e.yaml
+++ b/config/forecasters-co1e.yaml
@@ -4,8 +4,8 @@ description: |
   (KENDA-1) at 1km resolution.
 
 dates:
-  start: 2020-01-01T12:00
-  end: 2020-01-10T00:00
+  start: 2020-08-01T12:00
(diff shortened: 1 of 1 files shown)


config/forecasters-co1e.yaml
```diff
@@ -4,8 +4,8 @@ description: |
   (KENDA-1) at 1km resolution.
 
 dates:
-  start: 2020-01-01T12:00
-  end: 2020-01-10T00:00
+  start: 2020-08-01T12:00
+  end: 2020-08-10T00:00
```
dnerini: any specific reason to change the month in the example?

PR #73, opened 2025-10-21
Add make command to CLI

frazane:
Sometimes we just want to create a single file, for example when debugging. This new command allows to specify that single file as a target, and only run the workflow to generate that file. 

For example:

```shell
evalml make config/interpolators.yaml output/data/runs/8d1e0410c-d0846032f/verif_aggregated.nc
```

Gotcha: the file must be one that would be generated by the provided config (e.g. I cannot make a file from an interpolator using the `forecasters.yaml` config).

### DIFF
--- a/src/evalml/cli.py
+++ b/src/evalml/cli.py
@@ -128,9 +128,7 @@ def showcase(configfile, cores, verbose, dry_run, unlock, report, extra_smk_args
     )
 
 
-@cli.command(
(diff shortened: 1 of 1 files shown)


jonasbhend (approved):
Looking good thanks, just one suggestion. If make allows us to more easily debug, I think it would be useful to document this.

PR #74, opened 2025-10-22
Plot meteograms

dnerini:
This PR adds support for plotting meteograms within the evaluation framework. It introduces a new plotting pipeline that generates time-series visualizations of station observations and forecast output. The implementation integrates plotting logic with existing workflows and updates dependencies. It also brings in station observations from the PeakWeather dataset for richer observational context.

### DIFF
--- a/workflow/rules/plot.smk
+++ b/workflow/rules/plot.smk
@@ -33,7 +33,7 @@ rule plot_meteogram:
     resources:
         slurm_partition="postproc",
         cpus_per_task=1,
-        runtime="5m",
+        runtime="10m",
     params:
         grib_out_dir=lambda wc: (
             Path(OUT_ROOT) / f"data/runs/{wc.run_id}/{wc.init_time}/grib"


workflow/Snakefile
```diff
@@ -51,11 +51,18 @@ rule showcase_all:
...
             param=["2t", "10sp"],
             projection="orthographic",
             region=["none", "switzerland"],
         ),
+        expand(
+            OUT_ROOT / "showcases/{run_id}/meteograms/{init_time}_{param}_{sta}.png",
+            init_time=[t.strftime("%Y%m%d%H%M") for t in REFTIMES],
+            run_id=collect_all_candidates(),
+            param=["2t"],
+            sta=["GVE", "KLO", "LUG"],
```
jonasbhend: Parameters and stations should probably be added to the config.

jonasbhend: This comment still stands. I suggest to make parameters and stations configurable for the meteogram. Could also do this for the other showcase plots (i.e. use the same parameters).

workflow/rules/plot.smk
```diff
@@ -9,6 +9,52 @@ include: "common.smk"
...
+
+rule plot_meteogram:
+    input:
+        script="workflow/scripts/plot_meteogram.mo.py",
+        inference_okfile=rules.execute_inference.output.okfile,
+        analysis_zarr=config["analysis"].get("analysis_zarr"),
+        baseline_zarr=lambda wc: _use_first_baseline_zarr(wc),
+        peakweather_dir=rules.download_obs_from_peakweather.output.peakweather,
+    output:
+        OUT_ROOT / "showcases/{run_id}/{init_time}/{init_time}_{param}_{sta}.png",
```
jonasbhend: Is this a concious decision to produce meteogram plots per run_id? I can see that this ties in nicely with the showcases, but in this case, I would argue we do the meteograms only for an individual run (without a baseline). Otherwise, we should strive to support an arbitrary number of run_ids and baselines in the same (multi-panel) plot.

jonasbhend (commented):
Nice work, thanks Daniele. I have a few questions / comments and I suggest we discuss a bit more how we want to shape evalml going forward. To me what is not clear is how we organize the different commands/flavors (experiment and showcase). So far we have:

1. **experiment**: scores and metrics visualized for all forecasts and baselines jointly
2. **showcase**: example forecast visualizations (no observations)

The meteograms in this PR sit somewhere in between as they combine one run with one baseline, and I am not sure this is the best way to go. Instead I see two ways forward:

1. Align meteogram with current showcase, i.e. just a meteogram per run / baseline, no comparison (simple implementation)
2. Add support for an arbitrary number of runs and baselines in the same multi-panel plot (my favorite, but probably hard(er) to implement)

What do you think: @Louis-Frey @andreaspauling @dnerini @frazane

Louis-Frey (approved):
Thanks for this extension of evalml! Looks good to me.

workflow/Snakefile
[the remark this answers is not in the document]
dnerini: it's not yet supported in the interpolator

dnerini:
thank you all for the valuable feedback! as mentioned earlier, I suggest to merge the current status and incrementally improve on it with separate PRs

PR #80, opened 2025-11-14
Fix paths to baseline zarrs

dnerini:


config/interpolators.yaml
```diff
@@ -43,7 +43,7 @@ baselines:
   - baseline:
       baseline_id: COSMO-E-1h
       label: COSMO-E
-      root: /scratch/mch/bhendj/COSMO-E
+      root: /store_new/mch/msopr/ml/COSMO-E
```
jonasbhend:
```suggestion
      root: /store_new/mch/msopr/ml/COSMO-E_hourly
```

jonasbhend (approved):
Fixed path to hourly COSMO-E baseline dataset. Good to go

PR #90, opened 2025-12-11
MRB-665 Support for global models and remove meteodata-lab

frazane:
This PR introduces minimal support for global models in evalml, as we might want to evaluate small experiments on coarse global grids (e.g. o48 or o96).

**Summary of changes**
* optionally disable setting the `ECCODES_DEFINITION_PATH` environment variable when running global models 
* added base anemoi-inference config for global models and an example evalml config
* switch from meteodatalab to earthkit-data (with its xarray engine) to load GRIB files during verification 

TODO:
- [ ] Deal with the field `_earthkit` added to the xarray object by earthkit. New versions of xarray crash when to_netcdf with fields that are not allowed. Solution is to always `pop` it away. Temporary solution is a pin of xarray but its not sustainable.


dnerini:
experiment workflows works, e.g
image

I'm currently debugging the showcase workflow, but I suppose it'd make already sense to have another review round here.

dnerini:
@frazane can you also have a look pls? Supporting tp from the global GRIB files has been a bit of a pain and some of this fixes are really unpleasant ....

src/data_input/__init__.py
```diff
@@ -108,9 +135,43 @@ def load_fct_data_from_grib(
...
-    for var, da in ds.items():
+
+    profile = earthkit_xarray_engine_profile()
+    LOG.debug(f"loading GRIB for params {params} and steps {steps} from {root}")
+    # Extend param selection to include IFS aliases (e.g. "tp" for "TOT_PREC") so
+    # that both COSMO-named and IFS-named GRIB files (global models) are handled.
+    params_sel = list(
+        {p for p in params} | {_COSMO_TO_IFS[p] for p in params if p in _COSMO_TO_IFS}
+    )
+    # Precipitation params don't have a step=0 field (accumulation is zero at
```
dnerini, line 150: I suppose this was previously handled by meteodata-lab?

MicheleCattaneo:
Is it possible that there is an issue with the verification at specific locations?
202402010000_T_2M_KLO

This is T_2M. I understand that ERA is 0.25 degrees, but there is not even the diurnal cycle here. What do you think? Another issue, the lines dont start from the same point, but this model uses ERA5 as IC (unless I have an issue in my evalml config?)

src/data_input/__init__.py
jonasbhend: I guess this didn't occur previously because we haven't read from the global files at all. All our lam cutouts use COSMO names.

src/data_input/__init__.py
```diff
@@ -123,11 +138,28 @@ def load_fct_data_from_grib(
...
+    _PREC_PARAMS = {"tp", "TOT_PREC"}
+    prec_params = [p for p in params_sel if p in _PREC_PARAMS]
+    other_params = [p for p in params_sel if p not in _PREC_PARAMS]
+    fieldlist = ekd.from_source("file", files)
+    datasets = []
+    if other_params:
+        datasets.append(fieldlist.sel(param=other_params, step=steps).to_xarray(profile=profile))
+    if prec_params:
+        prec_steps = [s for s in steps if s > 0]
+        datasets.append(fieldlist.sel(param=prec_params, step=prec_steps).to_xarray(profile=profile))
```
jonasbhend: in `data_extract_baseline.py` we use the following pattern:

```
       for lt in lead_times:
            lh = lt % 24
            ld = lt // 24
            filepath = file / "grib" / f"{gribname}{ld:02}{lh:02}0000_{run_id}"
            LOG.info(f"Extracting {filepath}.")
            fields = ekd.from_source("file", filepath)
            for field in fields:
                if field.metadata("shortName") in params:
                    out.append(field)
```
That one looks slow to me, but I haven't profiled.

jonasbhend: The above approach has the advantage that you do not hard-code expectations in, in the sense that if there was a precipitation field at time 0 (with zero or NA in it), then you could still read it.

frazane:
In order to use earthkit-data v1.0 release candidates, we need up-to-date `eccodes` definitions that have not yet been published by DWD, but that we have internally. The CI is failing because of this as it cannot install the definitions from a private repository. 

@cosunae are there any updates from DWD?

MicheleCattaneo:
@frazane  the issue with the `_earthkit` field is still there with the updated dependencies, I assume?

cosunae:
DWD eccodes released on open data version 2.44 http://opendata.dwd.de/
we have a private copy of 2.47

jonasbhend:
I guess this PR is superseeded by #204 and could be closed, right?

dnerini:
Yes!

PR #92, opened 2026-01-07
MRB-650 Add score maps

jonasbhend:
# Opt-in score maps for runs and baselines

Adds a pipeline that produces temporally aggregated spatial verification maps (per-grid-point BIAS / RMSE / MAE / STDE) for both model runs and baselines. The computation is heavy, so it is opt-in via config.

## What's new

**Config**
- New optional `experiment.scoremaps` block controls the feature. Set `enabled: true` to produce score maps alongside the standard pipeline; omit the block or set `enabled: false` to skip it entirely (existing configs are unaffected).
- Fields: `params`, `leadtimes`, `scores`, `regions`, `seasons`, `init_hours`.
- `leadtimes` is validated at config-load time against every participant's forecast `steps`: requesting a lead time a run or baseline does not produce fails fast with a clear message (e.g. 36 h against an ICON-CH1 baseline with steps `0/33/6`). Accumulated params (TOT_PREC) require a lead time of at least one accumulation period.
- JSON schema regenerated; `ScoreMapsConfig` forbids unknown keys.
- _Replaces the earlier `--maps` CLI flag, which has been removed in favour of this config option._

**Workflow**
- Score-map targets are appended to `experiment_all` only when `experiment.scoremaps.enabled` is true.
- New rules `verification_scoremaps` (runs, GRIB input) and `verification_scoremaps_baseline` (baselines, ICON GRIB archive or INCA), plus `plot_scoremaps` / `plot_scoremaps_baseline` for the map plots.
- Score-map data: `data/{runs,baselines}/…/scoremaps/{param}_{leadtime}_{truth_hash}.nc`; plots: `results/{experiment}/scoremaps/{runs,baselines}/`.
- A single shared score colormap is used across RMSE/MAE/STDE for comparability.

**Script** (`workflow/scripts/verification_scoremaps.py`)
- Single unified script handling run (GRIB) and baseline (ICON GRIB archive / INCA NetCDF) inputs via mutually exclusive `--run_root` / `--baseline_root`.
- Streaming aggregation: the accumulators stream over init times, so no per-init-time error fields are written to disk.
- Computes BIAS/RMSE/MAE/STDE plus an N (sample-count) field, stratified by season (DJF/MAM/JJA/SON/all) and init hour (00/06/12/18/all).
- `--reftimes` restricts processing to the configured hindcast period (essential for baselines, whose archive is a continuous time series).
- TOT_PREC is de-accumulated over the `[lead − period, lead]` window; a missing step-0 field is synthesised as a zero initial condition when step 0 is requested.

## Testing

- Unit tests for the TOT_PREC de-accumulation / step-0 handling.
- Dry-run gating: feature off → no score-map jobs; on → correct job expansion; the leadtimes validator rejects unsupported lead times.
- Numerical regression against reference score maps: instantaneous params (T_2M, TD_2M) reproduce the references bit-identically (sample counts exact); TOT_PREC validated independently against a hand computation (de-accumulate → regrid → difference) to float32 precision.
- Full 2-init end-to-end run (verification → aggregation → score maps → plots → dashboard) completes cleanly.
- Dry-run sweep of all example configs, with and without score maps, builds correctly on the current tree (with `main` merged in).

## Deferred to follow-up PRs

- GRIB loader per-call overhead (TODO marker in `data_input/__init__.py`).
- Wind direction / vector visualisation.
- Reorganisation of pre-existing scalar plots under `plots/`.
- Further `data_input` consolidation (owned by the `refactor/data-io` branch).
- Nicer country polygons for map plots.
- Plot-every-pixel rendering

## Authors

Co-authored-by: Louis Frey <louis.frey@meteoswiss.ch>
Co-authored-by: Francesco Zanetta <francesco.zanetta@meteoswiss.ch>
Co-authored-by: Jonas Bhend <jonas.bhend@meteoswiss.ch>


src/plotting/__init__.py
```diff
@@ -29,12 +29,16 @@
         "extent": [-16.0, 25.0, 30.0, 65.0],
         "projection": _PROJECTIONS["orthographic"],
     },
+    # The domains which are originally called "centraleurope" and "switzerland"
+    # are mostly the same. I suggest making domain "switzerland" much smaller, 
+    # so that more spatial detail can be seen, especially in the complex 
+    # topography of the alps.
```
dnerini: it's a good idea! please move tihs comment to the description of your PR where you summarize the main changes introudced by this PR

src/plotting/colormap_defaults.py
```diff
@@ -44,6 +44,41 @@ def _fallback():
         "units": "mm",
         "levels": [0, 0.05, 0.1, 0.25, 0.5, 1, 1.5, 2, 3, 4, 5, 6, 7, 100],
     },
+
+    # hard-code this for the moment, can still make smarter later on:
```
dnerini: could you add these colormaps to the already existing ones? https://github.com/MeteoSwiss/evalml/blob/a42ce1ac3c19c793a4d40f58c01078a90698965c/src/plotting/colormap_defaults.py#L14

workflow/scripts/report_experiment_dashboard.py
```diff
@@ -28,19 +28,81 @@ def program_summary_log(args):
...
+    )
+
+    return df
+
 def main(args):
     program_summary_log(args)
 
     # Load, de-duplicate lead_time, and keep best provider per source (same logic as verif_plot_metrics)
-    dfs = [xr.open_dataset(f) for f in args.verif_files]
+    drop_variables = ["TOT_PREC.MAE.spatial", "TOT_PREC.RMSE.spatial", "TOT_PREC.BIAS.spatial",
```
dnerini: instead of hardcoding the variables names, why not just opening the first dataset in the `verif_files` list and extract all variable names including the keyword "spatial"?

workflow/scripts/report_experiment_dashboard.py
```diff
@@ -28,19 +28,81 @@ def program_summary_log(args):
...
-    nonspatial_vars = [d for d in ds.data_vars if "spatial" not in d]
-    df = ds[nonspatial_vars].to_array("stack").to_dataframe(name="value").reset_index()
+    # nonspatial_vars = [d for d in ds.data_vars if "spatial" not in d]
+    # df = ds[nonspatial_vars].to_array("stack").to_dataframe(name="value").reset_index()
+
+    df = dataset_metrics_to_dataframe(
+        ds,
+        forbidden_dims=("values",),  # critical!
+        metric_dims=("source", "season", "init_hour", "region", "lead_time", "eps"),
+    )
```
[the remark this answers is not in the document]
dnerini: then it'd be good to understand why that isn't behaving as expected and use only one single strategy to get rid of the spatial fields

src/plotting/colormap_defaults.py
dnerini: sorry, my bad, I was too quick, all good!

frazane (commented):
Beware you added some log files and some notebooks, I don't know if this was intended!

workflow/scripts/report_experiment_dashboard.py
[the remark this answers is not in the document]
Louis-Frey: These issues are circumvented now. The spatial information now never enters the `verif_aggregated.nc` files but instead is written to its own files. See commit ef22a3f80b924e00da27a86909aae9a1c2dd30d1.

Louis-Frey:
Fixed all issues remaining in my view and tested with the ICON example configs. After pinning the anemoi-inference version to 0.10.0, they run without error. The PR is ready. Please check again @dnerini @jonasbhend @frazane

.gitignore
dnerini, line 1: I suppose these additions should be removed?

src/plotting/__init__.py
```diff
@@ -29,12 +29,17 @@
         "extent": [-16.0, 25.0, 30.0, 65.0],
         "projection": _PROJECTIONS["orthographic"],
     },
+    # The domains which are originally called "centraleurope" and "switzerland"
+    # are mostly the same. I suggest making domain "switzerland" much smaller,
+    # so that more spatial detail can be seen, especially in the complex
+    # topography of the alps. Also, make domain "centraleurope" slightly smaller,
+    # to zoom in more on the LAM area, so that less plotting area outside is lost.
```
dnerini: imho comments about changes do not belong to the code, I'd rather propose such changes in the PR and if this is merged then it's adopted and there is no need to keep track of the motivation

workflow/rules/plot.smk
```diff
@@ -142,3 +142,48 @@ rule make_forecast_animation:
         """
         convert -delay {params.delay} -loop 0 {input} {output}
         """
+
+
+rule plot_metric_maps:
+    localrule: True
```
dnerini: sure this should be run locally?

workflow/scripts/verification_metrics.py
```diff
@@ -72,6 +72,14 @@ def main(args: ScriptConfig):
...
     truth = truth.sel(time=fcst.time)
 
+    # before verifying, calculate wind speed:
+    for ds in [fcst, truth]:
+        if "U_10M" in ds and "V_10M" in ds:
+            LOG.info("Calculating Wind Speed (SP_10M)...")
+            ds["SP_10M"] = (ds["U_10M"] ** 2 + ds["V_10M"] ** 2) ** 0.5
+            # Optional: Add metadata for the netCDF output
+            ds["SP_10M"].attrs = {"units": "m/s", "long_name": "10m Wind Speed"}
+
```
jonasbhend: I suggest to leave this out for now. Instead what we would want to do, is to configure the parameters for the experiments and to be able to (only) evaluate `SP_10M` for example.

workflow/scripts/verification_score_maps.py
```diff
@@ -0,0 +1,546 @@
...
+        return (ds["U_10M"] ** 2 + ds["V_10M"] ** 2) ** 0.5
+    raise ValueError(f"No recipe for derived variable '{param}'")
+
+
+# ---------------------------------------------------------------------------
+# Truth loading
+# ---------------------------------------------------------------------------
+
+
+def _open_zarr_component(root: Path, param: str) -> xr.DataArray:
```
jonasbhend: Is this functionality that is unique to the maps application or could this be introduced in `load_truth_data` (in fact in `load_analysis_from_zarr`)?

workflow/scripts/verification_score_maps.py
```diff
@@ -0,0 +1,546 @@
...
+        }
+        ds = xr.Dataset(components)
+        return _compute_derived(ds, param)
+    return _open_zarr_component(root, param)
+
+
+# ---------------------------------------------------------------------------
+# GRIB step helpers
+# ---------------------------------------------------------------------------
+
```
jonasbhend: missing or just not necessary any more?

.gitignore
Louis-Frey: Oops, yes, they eluded me...

workflow/scripts/verification_score_maps.py
```diff
@@ -0,0 +1,546 @@
...
+        default=None,
+        help="Root directory of a baseline (e.g. /path/to/ICON-CH1-EPS), containing FCST<YY>.zarr files.",
+    )
+    parser.add_argument(
+        "--baseline_zarrs",
+        type=Path,
+        nargs="+",
+        default=None,
+        help="Explicit list of baseline zarr paths (used by Snakemake for dependency tracking).",
+    )
```
jonasbhend: Do we really need both? baseline zarrs are not an output from a downstream rule, so baseline_root should be sufficient.

workflow/rules/plot.smk
Louis-Frey: Commented it, so it can be uncommented when interactive map plotting is desired in the future.

workflow/scripts/verification_score_maps.py
```diff
@@ -0,0 +1,546 @@
...
+    args = parser.parse_args()
+
+    if bool(args.run_root) == bool(args.baseline_root):
+        parser.error("Exactly one of --run_root or --baseline_root must be provided.")
+
+    if args.output is None:
+        source = args.run_root or args.baseline_root
+        args.output = (
+            source / f"verification_metric_maps_{args.param}_step{args.step:03d}h.nc"
+        )
```
jonasbhend: If I understand correctly, this would try to write into the baseline_zarr root (e.g. /store_new/mch/msopr/ml/ICON-CH1-EPS) if no output is provided. That would almost certainly fail or be not what we want to do. Instead, I would expect the maps NetCDF files to live under `output/data/runs/<run_id>/<netcdf>.nc` or `output/data/baselines/<baseline_id>/<netcdf>.nc`. I suggest to change this accordingly.

workflow/scripts/verification_metric_maps.py
```diff
@@ -0,0 +1,546 @@
...
+            f"{args.param}.BIAS": _seasonal_da(
+                lambda n, s: np.where(n > 0, accum_sum_e[s] / n, np.nan)
+            ),
+            f"{args.param}.RMSE": _seasonal_da(
+                lambda n, s: np.where(n > 0, np.sqrt(accum_sum_se[s] / n), np.nan)
+            ),
+            f"{args.param}.MAE": _seasonal_da(
+                lambda n, s: np.where(n > 0, accum_sum_ae[s] / n, np.nan)
+            ),
+            f"{args.param}.N": _seasonal_da(lambda n, s: np.where(n > 0, n, np.nan)),
```
jonasbhend: The components allow also to compute the standard deviation of error:

```
f"{args.param}.STDE": _seasonal_da(
    lambda n, s: np.where(n > 0, np.sqrt(accum_sum_se[s]/n - (accum_sum_e[s] / n)**2), np.nan)
),
```

workflow/scripts/verification_metrics.py
Louis-Frey: Makes sense. Removed it.

src/plotting/__init__.py
Louis-Frey: Removed it and also other similar comments.

jonasbhend (commented):
Hi @Louis-Frey. Wow this has come a long way, since I last had a look. Congrats!

I have a number of points, that I would like to discuss. More top-level ones are summarized below:

1. Naming
In the evaluation part, we use the terms scores and metrics, where metrics are independent of the ground truth (i.e. a 'statistic' of the forecast alone). This is not set in stone in the community, but I think it would help to keep the naming consistent at least throughout evalml.

2. Config
Instead of computing the scores for a specific lead time, I think we should be able to compute the scores for all the lead times as specified in the corresponding entries of the runs in the config. This would allow us to store results (.nc files) that are somewhat independent of the time slices selected for visualization. If the performance / time it takes to do compute all steps is really prohibitive, we should discuss it.

3. Precip-related code
Computing error components for all steps should also allow us to avoid issues with disaggregation (and the need to load adjacent time steps for accumulated quantities). I have the impression that there is quite a bunch of precip related code, that may become redundant.

4. Output paths
Currently the spatial scores are stored in files such as: `output/data/baselines/ICON-CH1-EPS/metric_maps/T_2M_24.nc`. I think at least for baselines, we need a different approach, as we run into problems when using the same baseline in different experiments (e.g. verification against analysis, verification against stations or low and high-resolution analysis data).

5. Adjacent work
There is a bunch of changes that are related to other ongoing work. Most notably probably the harmonization of data input across experiments and showcases. Therefore, I would strongly suggest to coordinate. What is not clear to me is, what functionality from the current data_input module is missing for this PR. If we manage to integrate all the necessary features in the data_input module, the harmonization is likely much easier.

@cosunae is working on a side-by-side visualization of maps for the showcases. Maybe this functionality could be leveraged to show side-by-side animations (plots) of maps of scores?

Addendum:
My heart bleeds when I see the manual iteration through reftime and summing up of components. I was really hoping we could avoid this (I know we can't currently). So kudos to you, @Louis-Frey, for implementing this nonetheless.

Louis-Frey:
Hi @jonasbhend, thanks a lot for the feedback and the kudos :-) I will address your points next week, now I have to leave soon. Regarding the 

> manual iteration through reftime and summing up

I have to give credit to @frazane too, it was mainly his idea.

workflow/scripts/verification_score_maps.py
Louis-Frey: Yes, ideally this should be handled in src/data_input. Since @clairemerker is refactoring data loading, I refrained from interfering and instead kept the code as-is, but added a TODO as a reminder to refactor later. See commit 7dfd8bea44fc71cac6d73e1db50954ab8e11d56a.

workflow/scripts/verification_score_maps.py
Louis-Frey: not necessary any more, removed. See 3cda7782fb856da9ff831470af6533d4cebd6cc4

workflow/scripts/verification_score_maps.py
Louis-Frey: Good catch — agreed the two were redundant. The script now globs FCST*.zarr under --baseline_root for init-time discovery, and I've dropped --baseline_zarrs entirely. Since the discovered init times get filtered down to the configured reftimes anyway, picking up extra archive years via the glob is harmless.

One nuance: I kept baseline_zarrs as an input: in the rule (just not passed on the CLI). Snakemake dependency tracking applies to all declared inputs, not only rule-produced ones — so listing them still gives existence checks and re-runs the rule if the baseline archive changes. The script simply doesn't need to receive the list to get that. Added a comment in the rule documenting this.

workflow/scripts/verification_score_maps.py
Louis-Frey: Good point. Fixed in 082ff376fa27631811a0593022b7afb2f479dcc7

workflow/scripts/verification_score_maps.py
jonasbhend: Unfortunately, there are no zarr archives for baselines anymore since #164

workflow/scripts/verification_metric_maps.py
Louis-Frey: Implemented. See 24ac031c78838d2396e547295be210670b2eb150

workflow/scripts/verification_score_maps.py
Louis-Frey: Thanks for the hint. I will wait for @clairemerker's `refactor/data-io` to be merged into main before refactoring.

Louis-Frey: I coordinated with Claire, and agreed it makes sense for my PR to land first. I merged main into this branch and am making required adjustments.

config/forecasters-ich1-oper-fixed.yaml
```diff
@@ -85,3 +85,36 @@ profile:
     runtime: "1h"
     gpus: 0
   jobs: 50
+
+score_maps:
```
dnerini: this should go under the `experiment:` section, similarly to `score_cards`, right?

src/evalml/cli.py
```diff
@@ -163,9 +164,24 @@ def cli():
 @click.argument(
     "configfile", type=click.Path(exists=True, dir_okay=False, path_type=Path)
 )
+@click.option(
+    "--maps",
+    is_flag=True,
+    default=False,
+    help="Also produce score maps (computationally intensive).",
+)
```
dnerini: instead of an ad-hoc flag, elsewhere we used an "enabled" parameter in the config, see e.g https://github.com/MeteoSwiss/evalml/blob/8a650c8d2bb875b63e737fb64cd5dda58f2edddb/src/evalml/config.py#L300

workflow/Snakefile
```diff
@@ -160,6 +160,33 @@ rule experiment_all:
         ),
 
 
+rule score_maps_all:
```
dnerini: normally we would include the score maps output in the experiment_all rule, or is there a particular reason to have it separately?

dnerini (changes requested):
I left some minor comments. The PR is looking really good!

Louis-Frey:
Hi @dnerini, thanks for the comments. I will try to incorporate them and get the PR ready to merge asap.

workflow/scripts/verification_score_maps.py
Louis-Frey: Addressed in commit ed56fd4de1421d040713a2abc7b8b132c9458a79. Marking as resolved.

config/forecasters-ich1-oper-fixed.yaml
Louis-Frey: addressed in 4d8fe7e076efd3341e753233f918a12957883bac.

Louis-Frey:
Hi @jonasbhend, extensive answer to your comments:

> Hi @Louis-Frey. Wow this has come a long way, since I last had a look. Congrats!
> 
> I have a number of points, that I would like to discuss. More top-level ones are summarized below:
> 
> 1. Naming
>    In the evaluation part, we use the terms scores and metrics, where metrics are independent of the ground truth (i.e. a 'statistic' of the forecast alone). This is not set in stone in the community, but I think it would help to keep the naming consistent at least throughout evalml.
> 2. Config
>    Instead of computing the scores for a specific lead time, I think we should be able to compute the scores for all the lead times as specified in the corresponding entries of the runs in the config. This would allow us to store results (.nc files) that are somewhat independent of the time slices selected for visualization. If the performance / time it takes to do compute all steps is really prohibitive, we should discuss it.
> 3. Precip-related code
>    Computing error components for all steps should also allow us to avoid issues with disaggregation (and the need to load adjacent time steps for accumulated quantities). I have the impression that there is quite a bunch of precip related code, that may become redundant.
> 4. Output paths
>    Currently the spatial scores are stored in files such as: `output/data/baselines/ICON-CH1-EPS/metric_maps/T_2M_24.nc`. I think at least for baselines, we need a different approach, as we run into problems when using the same baseline in different experiments (e.g. verification against analysis, verification against stations or low and high-resolution analysis data).
> 5. Adjacent work
>    There is a bunch of changes that are related to other ongoing work. Most notably probably the harmonization of data input across experiments and showcases. Therefore, I would strongly suggest to coordinate. What is not clear to me is, what functionality from the current data_input module is missing for this PR. If we manage to integrate all the necessary features in the data_input module, the harmonization is likely much easier.
> 
> @cosunae is working on a side-by-side visualization of maps for the showcases. Maybe this functionality could be leveraged to show side-by-side animations (plots) of maps of scores?
> 
> Addendum: My heart bleeds when I see the manual iteration through reftime and summing up of components. I was really hoping we could avoid this (I know we can't currently). So kudos to you, @Louis-Frey, for implementing this nonetheless.

1. Naming: Addressed in e9fbef94a100e805f1e723cc4f6aab8695bd9fab and 91983fe5fe3bd46f6f8b9bd55cfe12492bbfaa8e

2. Config: As discussed, I left it at processing individual lead times, due to memory and disk use constraints, and because we anticipate temporal granularity to increase substantially with temporal downscaler and nowcasting. 

3. Precip-related code: Since this comment, we ported score-maps loading to the shared post-earthkit data_input, which deleted the script's own de-accumulation shim — disaggregation now lives in one place (_tot_prec_handling, on main, shared with point verification) and contributes no code to this PR. What remains precip-specific here is ~40 lines: truth-variable selection (TOT_PREC_<N>H), units convention, and the accumulation-window derivation. Computing all steps would remove only the last (~25 lines), while the memory/disk constraints discussed above still apply, and per-step error components fundamentally can't reconstruct RMSE/STDE of multi-hour accumulations — only BIAS is linear. The pad(0) code flagged earlier was removed on main by the earthkit migration; its replacement zero-fills only a present-but-all-NaN step 0 and raises on already-disaggregated input.

4. Output Paths: Addressed for baselines in eb0f5195670b8b4dfb3219ed84f047eaeba3f944 and 7173cc22af8f31014072afeeaae07814d900d3a3

5. Coordinated with @clairemerker. Only very minor change left in src/data_input/__init__.py

src/evalml/cli.py
Louis-Frey: Addressed in c941695caa5427aa77ec8ed8452807baaec909c0. Implemented the "enabled" parameter (default is "not enabled").

workflow/Snakefile
Louis-Frey: Good point. It is now also addressed in c941695caa5427aa77ec8ed8452807baaec909c0.

Louis-Frey:
Thoroughly testing the PR now.

Louis-Frey:
Update:

Since the last round I ran a full regression test of the score-maps path (instantaneous params reproduce the pre-merge reference maps bit-identically; TOT_PREC validated end-to-end against an independent hand computation). It surfaced four bugs, now fixed:

af061b9 — TOT_PREC step-0 handling: the first-lead-time [0, period] window crashed (and full-range loads silently returned all-NaN precip at the first lead time).
3fe2531 — run-side score-maps rule was missing the ECCODES definitions export (empty maps) and hardcoded output/ instead of the configured output_root.
780cae8 — score-map plotting broke after the earthkit/loader migration (coord rename + Colormap-object rejection).
6e7d591 — score-map lead times are now resolved per participant: a 36h map is no longer scheduled for an ICON-CH1 baseline (steps 0/33/6), and leadtimes: all no longer plans doomed jobs.

The PR is ready in my view. Please check again @dnerini @jonasbhend

workflow/rules/verification.smk
```diff
@@ -169,3 +169,82 @@ rule verification_metrics_plot:
...
+    input:
+        "src/verification/__init__.py",
+        "src/data_input/__init__.py",
+        script="workflow/scripts/verification_score_maps.py",
+        forecast=lambda wc: BASELINE_CONFIGS[wc.baseline_id]["root"],
+        truth=config["truth"]["root"],
+        eckit_grids=rules.data_download_eckit_geo_grids.output,
+    output:
+        OUT_ROOT
+        / f"data/baselines/{{baseline_id}}/{config['truth']['label']}/score_maps/{{param}}_{{leadtime}}.nc",
```
jonasbhend: This output includes the truth label, instead I suggest to use the recently introduced truth_hash to make sure that results are truth specific.

workflow/rules/verification.smk
```diff
@@ -169,3 +169,82 @@ rule verification_metrics_plot:
...
+        "src/data_input/__init__.py",
+        script="workflow/scripts/verification_score_maps.py",
+        inference_okfiles=lambda wc: expand(
+            rules.inference_execute.output.okfile,
+            init_time=_restrict_reftimes_to_hours(REFTIMES),
+            allow_missing=True,
+        ),
+        truth=config["truth"]["root"],
+    output:
+        OUT_ROOT / "data/runs/{run_id}/score_maps/{param}_{leadtime}.nc",
```
jonasbhend: output and logs here should also include reference to truth hash (see below). To make sure we don't overwrite when switching the truth (which is less of a problem as we cannot do the maps with stations).

src/evalml/config.py
```diff
@@ -380,6 +423,10 @@ class ExperimentConfig(BaseModel):
         default=None,
         description="Scorecard generation configuration. Omit or set enabled: false to disable.",
     )
+    score_maps: ScoreMapsConfig = Field(
+        default_factory=ScoreMapsConfig,
+        description="Score map plot configuration. Set enabled: true to produce score maps.",
```
dnerini: I'd try to improve consistency with the existing fields, ie
- naming: `scoremaps` instead of `score_maps` (same as `scorecards`)
- default: `None` instead of `ScoreMapsConfig` (to give the option to omit it the field from the config)

workflow/Snakefile
```diff
@@ -158,6 +202,7 @@ rule experiment_all:
             )
             else []
         ),
+        (rules.score_maps_all.input if SCORE_MAPS_CONFIG["enabled"] else []),
```
dnerini: same as above, rename to `SCOREMAPS_CONFIGS`

workflow/Snakefile
```diff
@@ -130,6 +130,50 @@ onerror:
 # -----------------------------------------------------
 
 
+SCORE_MAPS_CONFIG = config["experiment"]["score_maps"]
+
+
+def _score_map_targets():
```
dnerini: I wonder if this can be simplified somehow? do we really need 40+ lines of code to expand on all the score maps?

dnerini (changes requested):
Nice work, thanks! A few more comments from my side.

src/data_input/__init__.py
```diff
@@ -232,6 +265,12 @@ def _tot_prec_handling(tp: xr.DataArray) -> xr.DataArray:
         "Disaggregating TOT_PREC from cumulative-from-start to per-step accumulations."
     )
     tp = tp.diff("step")
+    if tp.sizes["step"] == 0:
+        raise ValueError(
+            "Cannot de-accumulate TOT_PREC: only a single step was loaded and "
+            "step 0 was not requested/synthesised, so no accumulation window "
+            "can be formed. Request the preceding step as well."
+        )
```
jonasbhend: Shouldn't this be raised before trying to compute the difference?

src/data_input/__init__.py
jonasbhend, line 1: I think this needs some design choices. Currently the assumption is that grib-based forecasts (icon archive and ml foercasts) are in one file per leadtime. The current implementation for the maps PR, however, adds an additional argument to retrieve specific steps (not by selecting the files). I haven't completely grasped it, but I think there is some duplication. If we can safely assume that we can select lead times by selecting grib files, we should do this throughout. Maybe it would be sufficient to just call the higher-level function `load_forecast_data` to handle the steps abstraction?

workflow/Snakefile
dnerini: @Louis-Frey I gave it a try in 509808e, can you have a look?

dnerini: and f38c2ee, consequently

dnerini: and finally a validator to avoid requesting lead times that are not available 202b651

Louis-Frey: With this change, you are (at least partially) effectively reverting 6e7d5915e93a348b15932faa7d06aff527bdece3, is that deliberate?

dnerini: mmmh yes, the logic is now a bit simpler:

- always an explicit list of leadtimes (no "all" shortcut)
- same leadtimes for all models (the config validation will complain otherwise)

... too much simplification?

Louis-Frey: Hmm... Not sure if it's too much simplification, but the "all" option was deliberately introduced in 77c9a64bb25053eff8e3e8b0fef955025f8947e1, to satisfy the request of @jonasbhend in his comment of 8 May: 

"Instead of computing the scores for a specific lead time, I think we should be able to compute the scores for all the lead times as specified in the corresponding entries of the runs in the config. This would allow us to store results (.nc files) that are somewhat independent of the time slices selected for visualization. If the performance / time it takes to do compute all steps is really prohibitive, we should discuss it." 

I had a discussion with Jonas back then and we agreed it is best to implement the "all" option as syntactic sugar, as it is now.

Not sure if we should revert it now. I guess we could live with not having an "all" option, but switching back to an explicit list does not solve the problem that lead times can be requested from a participant which does not have these lead times (e.g., lead time >33h from ICON-CH1). But I guess you solved that in 202b651ef3917a1a7750be96e0d62c41507dbc10. Not sure what is better now. 

One argument to have an "all" option is that if we ever want to show animations of score maps with e.g. hourly resolution, then typing in a long list could be tedious (or can that be alleviated with a np.linspace() or similar?)

jonasbhend: I'm happy to revert. I don't remember the details exactly, but the current setup would only recompute missing lead times, so it's fine to not have the all list and only compute what is necessary at each invocation.

Louis-Frey: Also, my agent says:

"Heads up — I have local changes addressing the two consistency comments above (score_maps → scoremaps, default=None), and I've just pulled your three commits (509808e, f38c2ee, 202b651). Before I reconcile them I want to align on two things, because they interact:

1. The new validator vs. the default=None request. validate_score_map_leadtimes reads self.experiment.score_maps.enabled directly, which assumes the block is always present. That's in tension with dnerini's comment asking for default=None (omittable). If we go with None, the validator needs to guard for scoremaps is None and use the renamed field. I'm happy to fold the rename + None-guard into your commits — just confirm you want scoremaps/None and I'll rebase my naming changes on top of yours.

2. Baselines aren't validated or filtered for producible lead times. validate_score_map_leadtimes iterates self.runs only, and the inlined expansion in experiment_all expands plot_score_maps_baseline over every requested leadtime for every baseline. So a lead time that a run produces but a baseline doesn't (e.g. 36 h against the ICON-CH1 baseline, steps 0/33/6) passes validation and then schedules a baseline plot job with no input — which is exactly the case 6e7d591 (per-participant resolve_leadtimes) was added to prevent. Was dropping the per-participant resolution intentional? If we keep the hard-fail-validation approach, should it also cover baselines — or should baseline lead times still be filtered to what each baseline produces?

Once we settle these I'll push the reconciled version."

jonasbhend (commented):
Hi @Louis-Frey, here are a few more comments on the open PR. I have also taken it for a spin, to get a better understanding of what is happening, still have to look a bit more closely into results (filenames, rules, configs and such). Let me know if you need help / clarification.

src/data_input/__init__.py
```diff
@@ -212,13 +218,40 @@ def fieldlist_to_xarray(fieldlist) -> xr.Dataset:
...
+    if requested_steps is not None:
+        if 0 in requested_steps:
+            step0_idx = np.where(tp["step"].values == np.timedelta64(0, "ns"))[0]
+            if step0_idx.size == 0:
+                LOG.warning(
+                    "Step 0 of TOT_PREC is missing from the GRIB, prepending "
+                    "zeroes assuming accumulate_from_start_of_forecast is "
+                    "enabled."
+                )
+                zero = xr.zeros_like(tp.isel(step=[0]))
```
jonasbhend, line 243:
```suggestion
                zero = xr.zeros_like(tp.isel(step=0))
```

workflow/rules/verification.smk
```diff
@@ -169,3 +169,82 @@ rule verification_metrics_plot:
...
+    log:
+        OUT_ROOT / "logs/verification_score_maps/{run_id}-{param}-{leadtime}.log",
+    resources:
+        cpus_per_task=24,
```
jonasbhend: do we really need that many cpus?

workflow/Snakefile
Louis-Frey: Any comments, @dnerini? I think we need to settle this before I can move on.

Louis-Frey: Had a discussion with @dnerini. Issue is settled: We sticked to Daniele's version, with the naming adjustments from the above comments on top. See d91096c40af73bc9c16d517e52e8408eb4863804.

workflow/rules/verification.smk
Louis-Frey: Reduced to 2 in 17c527108354757ecb12b9258054744cc9da7087. See commit message for details.

src/data_input/__init__.py
Louis-Frey: Thanks — but I'd prefer to keep `isel(step=[0])` here. The list form is intentional: it keeps step as a length-1 dimension, which the next two lines rely on. We re-label that dim to step 0 via `zero.assign_coords(step=[np.timedelta64(0, "ns")])` and then `xr.concat([zero, tp], dim="step")`.

The scalar form `isel(step=0)` drops step to a scalar coordinate, so the `assign_coords` immediately raises:

```
CoordinateValidationError: coordinate step has dimensions ('step',), but these are not a subset of the DataArray dimensions ('values',)
```

The scalar equivalent would need an extra `expand_dims("step")` to put the dimension back, which is more lines for the same result. Happy to add a brief comment explaining why it's `[0]` if that helps readability.

workflow/rules/verification.smk
Louis-Frey: Addressed in d967250caea6ff6ded4fe3b49ded1b46700d2cd6. I deliberately used filename suffixes instead of directory paths and adjusted the run side as well to mirror main exactly.

workflow/rules/verification.smk
Louis-Frey: Addressed in 847f2d21695a479466a8e3c9430bf6569e6c793b. The log paths now also use `TRUTH_HASH`, completing the request.

src/evalml/config.py
```diff
@@ -219,6 +219,45 @@ class BaselineItem(BaseModel):
     baseline: BaselineConfig
 
 
+class ScoreMapsConfig(BaseModel):
```
jonasbhend, line 222: The config is not validated in the sense that
```
     scoremaps:
         metrics: ["BIAS"] # instead of scores
```
doesn't trigger an error.

workflow/scripts/verification_scoremaps.py
```diff
@@ -0,0 +1,676 @@
...
+        LOG.info("Using %d baseline init times from --reftimes", len(init_items))
+    else:
+        init_items = iter_init_dirs(args.run_root)
+        LOG.info("Found %d init time directories", len(init_items))
+
+        # Restrict to the experiment's configured init times if provided.
+        if args.reftimes:
+            wanted = {datetime.strptime(s, DATETIME_FMT) for s in args.reftimes}
+            init_items = [(rt, d) for rt, d in init_items if rt in wanted]
+            LOG.info("Filtered to %d init times matching --reftimes", len(init_items))
```
jonasbhend: Elsewhere in evalml we fail if not all initializations are available. Otherwise results from runs and baselines may not be comparable.

src/plotting/colormap_defaults.py
jonasbhend, line 1: ok, found the problem. Colourmaps for STDE are missing, so STDE is plotted with the 'standard' colormap for BIAS. I propose to adopt the alternative and fall back to the parameter-specific colormap for scores instead, that way you only have to define this once as a fallback (for MAE, RMSE and STDE the same) and then define the special case of BIAS.

src/data_input/__init__.py
Louis-Frey: Addressed in d522edfc81dd580b4c47e1906c4cdf631ed640b0. Marking as resolved.

src/data_input/__init__.py
Louis-Frey: Turns out the "duplication" of files and steps was actually necessary. See 03b24e6da4ad413ba337158d9e7a01731a80f27e. Changed only the docstring of `load_forecast_data_from_grib` to clarify.

src/evalml/config.py
Louis-Frey: Good catch - ScoreMapsConfig was missing the `extra: forbid` that the other config models use. Added it, so `metrics:` (or any unknown key) now raises `extra_forbidden` instead of silently falling back to the scores default. Verified with a quick check. See 7f049a8605207e649d4f25ff05c8a12fa1242122.

src/plotting/colormap_defaults.py
Louis-Frey: Agreed and addressed in 757cd893cc16b7d73cb2589518f5e4eea74c2af6.

workflow/scripts/verification_scoremaps.py
Louis-Frey: Addressed in 078c9683314edf0e84eec1b24a5bbce975e70e98. EvalML fails now if any of runs or baselines has a missing init, or if for some reason a valid time is missing from the truth.

Louis-Frey:
All comments raised before are addressed. Please check again @dnerini @jonasbhend. If there are no further issues / comments raised, I suggest you can proceed to testing, @jonasbhend. After you complete, I will re-run my comprehensive test engine and merge if nothing surfaces.

jonasbhend:
> All comments raised before are addressed. Please check again @dnerini @jonasbhend. If there are no further issues / comments raised, I suggest you can proceed to testing, @jonasbhend. After you complete, I will re-run my comprehensive test engine and merge if nothing surfaces.

Good for me, I tested it. Does what I expect.

dnerini (approved):
looking good for me too, thanks @Louis-Frey for the hard work!

Louis-Frey:
Good. Comprehensive tests complete and passing. Updated PR description. Merging.

PR #96, opened 2026-01-19
Disentangled forecasters experiments

frazane:
Adds new configs for running experiments with disentangled forecasters.


icedoom888 (commented):
✅

PR #100, opened 2026-01-26
Fix ICON-CH1 evaluation with baseline

frazane:


jonasbhend (approved):
i have tried it, it works and fixes the problem with the stratification. Amazing. Thanks a lot, @frazane

PR #101, opened 2026-01-26
Pin dependencies for COSMO-2 emulators examples

frazane:
Some of our older checkpoints, notably those trained on COSMO-2 analysis data, required python 3.10 to be installed. However, support for python 3.10 was recently dropped in anemoi-inference: https://github.com/ecmwf/anemoi-inference/pull/415

In this PR we pin anemoi-inference to v0.8.3 in example configs that use COSMO-2 emulators. The `showcase.yaml` config is removed because it was redundant (related: we will revisit the config/command interface of evalml at some point).


jonasbhend (approved):
Haven't rerun the example configs to make sure this works. @Louis-Frey if you have the time, could you please check?

Louis-Frey (approved):
Thanks Francesco! As far as I can tell, this looks good.

Wait a moment with merging, I will check the configs too (see Jonas' comment). Sorry a bit fast.

PR #110, opened 2026-02-12
Integration of MEC workflow

andreaspauling:
Add the MEC workflow. The new parts are in green in the DAG: [snakemake_dag.pdf](https://github.com/user-attachments/files/25264856/snakemake_dag.pdf)

For each valid date a MEC case is set up and run. This includes:
- creating the directory structure
- adding the observations
- organizing the model input including past runs depending on the config
- rendering the MEC namelist
- executing MEC for all dates with complete data for all leadtimes (excludes the first ones of the period)
- storing the final feedback file in a separate place.

All MEC cases can be removed once the final feedback file is produced (removal not yet implemented).

- Topics already raised by Francesco:
  - put folder mec/ in data/mec in order not to mix up init and valid time (MEC is valid time oriented)
  - check globbing options in MEC namelist with DWD (not documented, only _FCR_TIME_ is supported afaik, * etc not). The aim is to avoid copying data.


workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,206 @@
...
+        times.append(t.strftime("%Y%m%d%H%M"))
+        t += freq_td
+    return times
+
+
+# prepare_mec_input: setup run dir, gather observations and model data in the run dir for the actual init time
+rule prepare_mec_input:
+    input:
+        src_dir=OUT_ROOT / "data/runs/{run_id}/{init_time}/grib",
+        inference_ok=OUT_ROOT / f"run_inference_all.{EXPERIMENT_HASH}.ok",
```
dnerini:
```suggestion
        inference_okfile=rules.execute_inference.output.okfile,,
```

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,206 @@
...
+    while t <= base:
+        times.append(t.strftime("%Y%m%d%H%M"))
+        t += freq_td
+    return times
+
+
+# prepare_mec_input: setup run dir, gather observations and model data in the run dir for the actual init time
+rule prepare_mec_input:
+    input:
+        src_dir=OUT_ROOT / "data/runs/{run_id}/{init_time}/grib",
```
dnerini: There is no rule giving this an output, so this should to the very list trigger some warnings from snakemake. You could specify it as a parameter instead.

dnerini (commented):
looking very nice @andreaspauling ! I have added a few initial thoughts form a quick look into your changes, but I plan to have a closer look soon!

frazane (changes requested):
Added some comments. The most important ones are those we already discussed (avoid copying data and directory structure) but I wanted to make sure we do not forget about them.

workflow/scripts/generate_mec_namelist.py
```diff
@@ -0,0 +1,71 @@
...
+
+    parser.add_argument(
+        "--template",
+        type=str,
+    )
+
+    parser.add_argument(
+        "--namelist",
+        type=str,
+        help="Anything useful",
```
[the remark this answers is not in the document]
andreaspauling: Done

workflow/scripts/generate_mec_namelist.py
```diff
@@ -0,0 +1,71 @@
...
+    out_path = Path(str(args.namelist))
+    out_path.parent.mkdir(parents=True, exist_ok=True)
+    with out_path.open("w", encoding="utf-8") as f:
+        f.write(namelist)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+
+    parser.add_argument("--steps", type=_parse_steps, default="0/120/6")
```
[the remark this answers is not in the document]
andreaspauling: done

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,213 @@
...
+
+        # concatenate all grib files in src_dir into a single file fc_file
+        echo "grib files processed:"
+        files=( "$src_dir"/20*.grib )
+        if (( ${{#files[@]}} )); then
+            printf '%s\n' "${{files[@]}}"
+            cat "${{files[@]}}" > "$fc_file"
+        else
+            echo "WARNING: no grib files found in $src_dir" >&2
+        fi
```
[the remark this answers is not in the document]
andreaspauling: Actually, MEC does need all necessary fields for calculating precipitation (differences) in the same file. Now only the necessary files are put together (instead of all). This reduces the temporary disk space considerable. I will add also an option to delete all MEC cases on success, avoiding duplicating output data in the end.

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,213 @@
...
+    while t <= end:
+        times.append(t)
+        t += freq
+    return times
+
+
+REFTIMES_MEC = _reftimes_mec()
+
+
+def init_times_for_mec(wc):
```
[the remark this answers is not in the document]
andreaspauling: Lists are also supported now

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,213 @@
...
+        if (( ${{#files[@]}} )); then
+            printf '%s\n' "${{files[@]}}"
+            cat "${{files[@]}}" > "$fc_file"
+        else
+            echo "WARNING: no grib files found in $src_dir" >&2
+        fi
+
+        # collect observations (ekfSYNOP) and/or (monSYNOP from DWD; includes precip) files
+        cp /store_new/mch/msopr/osm/KENDA-1/EKF/${{ym}}/ekfSYNOP_${{init}}00.nc {output.ekf_file}
+        cp /scratch/mch/paa/mec/MEC_ML_input/monFiles2020/hpc/uwork/swahl/temp/feedback/monSYNOP.${{init:0:10}} {output.obs}/monSYNOP.nc
```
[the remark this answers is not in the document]
andreaspauling: All paths are moved to the new example config for MEC/FFV2

workflow/rules/verif_obs.smk
andreaspauling: This input was removed

workflow/Snakefile
```diff
@@ -36,6 +35,8 @@ DATA_DIR = OUT_ROOT / "data"
 LOGS_DIR = OUT_ROOT / "logs"
 RESULTS_DIR = OUT_ROOT / "results" / EXPERIMENT_NAME
 
+# prefer one rule because snakemake complains about ambiguous rules (same output)
+ruleorder: prepare_inference_forecaster > prepare_inference_interpolator
```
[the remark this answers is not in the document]
andreaspauling: It seems to be necessary and fails without. Claudes take:
Both rules produce the same three outputs:

data/runs/{run_id}/{init_time}/config.yaml
data/runs/{run_id}/{init_time}/resources/
data/runs/{run_id}/{init_time}/grib/
inference_prepare_interpolator adds a fourth output (forecaster/ directory) and has an extra input (the upstream forecaster's okfile), but the three shared outputs are identical. Snakemake cannot distinguish which rule to use for a forecaster run_id vs. an interpolator run_id based on the output pattern alone — it needs ruleorder to resolve the ambiguity by preferring inference_prepare_forecaster.

So the ruleorder is still needed and correct.

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,213 @@
...
+    # check that steps is in the format "start/stop/step"
+    if "/" not in steps:
+        raise ValueError(f"Expected steps in format 'start/stop/step', got '{steps}'")
+    if len(steps.split("/")) != 3:
+        raise ValueError(f"Expected steps in format 'start/stop/step', got '{steps}'")
+    start, end, step = map(int, steps.split("/"))
+    return list(range(start, end + 1, step))
+
+
+# TODO: merge with _ref_times from common.smk?
```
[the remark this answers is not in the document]
andreaspauling: Most probably MEC will stay the only use case. I would leave it here but wouldnt mind to move.

workflow/scripts/generate_mec_namelist.py
```diff
@@ -0,0 +1,71 @@
...
+        "--init_time",
+        type=lambda s: datetime.strptime(s, "%Y%m%d%H%M"),
+        default="202010010000",
+        help="Valid time for the data in ISO format.",
+    )
+
+    parser.add_argument(
+        "--template",
+        type=str,
+    )
```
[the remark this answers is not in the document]
andreaspauling: Done

workflow/rules/verif_obs.smk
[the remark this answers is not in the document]
andreaspauling: Now MEC runs outside the forecast run directory in a separate directory structure:  `output/data/runs/<run-id>/mec/<valid-time>`

andreaspauling:
- FFV2 in this PR as well
- evalml options --mec --ffv2 added (default: no mec/ffv2)
- support of lists in config
- mec running outside forecast run directories
- support ver-files as source for observations
- Paths moved to config
- minor fixes / cleanup

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,206 @@
...
+        fdbk_file=OUT_ROOT / "data/runs/{run_id}/fdbk_files/verSYNOP_{init_time}.nc",
+    log:
+        OUT_ROOT / "data/runs/{run_id}/{init_time}/mec/run_mec.log",
+    shell:
+        """
+        (
+        set -euo pipefail
+ 
+        # Run MEC inside sarus container
+        # Note: pull command currently needed only once to download the container
```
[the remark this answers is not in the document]
frazane: I second this.

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
frazane, line 1: Are these changes to the accumulation logic for total precipitation needed here? If not, I would remove these.

workflow/rules/inference.smk
```diff
@@ -196,9 +196,7 @@ rule inference_prepare_forecaster:
         config=Path(OUT_ROOT / "data/runs/{run_id}/{init_time}/config.yaml"),
         resources=directory(OUT_ROOT / "data/runs/{run_id}/{init_time}/resources"),
         grib_out_dir=directory(OUT_ROOT / "data/runs/{run_id}/{init_time}/grib"),
-        okfile=touch(
```
frazane, line 194: Why was this change made?

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,353 @@
+from pathlib import Path
+from datetime import timedelta
+
+
+def _parse_steps(steps: str) -> list[int]:
```
frazane, line 6: Isn't this a duplicate of https://github.com/MeteoSwiss/evalml/blob/e4af0a6a61329172f25a9a19659546b0b655280d/workflow/rules/plot.smk#L119?

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,353 @@
...
+        OUT_ROOT / "logs/prepare_mec_input/{run_id}-{init_time}.log",
+    params:
+        ekf_root=config["mec"]["ekf_root"],
+        mon_synop_root=config["mec"]["mon_synop_root"],
+        ver_synop_root=config["mec"]["ver_synop_root"],
+    shell:
+        """
+        (
+        set -euo pipefail
+        shopt -s nullglob
```
frazane: Why do we need this?

workflow/scripts/generate_ffv2_namelist.py
```diff
@@ -0,0 +1,148 @@
...
+def _make_veri_ens_member(experiment_ids: str) -> str:
+    num_ids = len(experiment_ids.split(","))
+    return ",".join(["-1"] * num_ids)
+
+
+def main(args):
+    # Render template with provided args
+    context = {
+        "experiment_ids": args.experiment_ids,
+        "feedback_directories": args.feedback_directories,
```
frazane: This is not used in the Jinja template as far as I can tell.

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,353 @@
...
+    log:
+        OUT_ROOT / "logs/run_mec/{run_id}-{init_time}.log",
+    shell:
+        """
+        (
+        set -euo pipefail
+
+        # Run MEC inside sarus container
+        # Note: pull command currently needed only once to download the container
+        sarus pull container-registry.meteoswiss.ch/mecctr/mec-container:0.1.0-main
```
frazane: As already pointed out, this should be refactored into a preparatory rule.

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,353 @@
...
+        cp {params.ekf_root}/${{ym}}/ekfSYNOP_${{init}}00.nc {output.ekf_file}
+        cp {params.mon_synop_root}/${{init:0:10}}/monSYNOP.nc {output.obs}/monSYNOP.nc
+        cp {params.ver_synop_root}/verSYNOP_${{init}}00.nc {output.obs_file}
+        echo "Copied obs files to {output.obs}"
+
+        ) > {log} 2>&1
+        """
+
+
+# link_mec_input: create the input_mod dir with symlinks to all fc files from all source inits
```
frazane: This rule is not creating symlinks, but copies. Didn't we want to avoid this?

workflow/rules/verif_obs.smk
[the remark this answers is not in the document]
andreaspauling: Done. Sorry, I overlooked this.

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
andreaspauling: Exactly. MEC needs precip accumulated from the beginning of the run

workflow/rules/inference.smk
andreaspauling: With Claudes help: The `okfile` is necessary. `inference_execute` needs to depend on whichever of the two prepare rules ran, but it can't reference them directly by output path because both produce the same three outputs (`config.yaml, resources/, grib/`). The `_inference_routing_fn` function selects the correct prepare rule by model type — but to do so, it must reference a path that is unique per rule. The okfile provides that.

That okfile is used in `_inference_routing_fn` . The routing function returns the okfile path of whichever prepare rule ran (forecaster or interpolator), and `inference_execute` declares it as its input — this is how Snakemake knows to wait for the correct prepare rule to finish before launching inference.

Sounds plausible to me.

frazane: What I mean is that `touch("/some/file")` already automatically generates the file when the rule succeeds.

workflow/rules/verif_obs.smk
andreaspauling: They have different input and output. It may be possible to merge but that would need some time and result in a one more complicated function.

workflow/rules/verif_obs.smk
andreaspauling: No. I removed it.

workflow/scripts/generate_ffv2_namelist.py
andreaspauling: You are right. feedback_directories and output_directory is legacy. Is removed.

workflow/rules/verif_obs.smk
andreaspauling: Agree. Done.

workflow/rules/verif_obs.smk
andreaspauling: I implemented a version with only symlinks. However, this did not work because all fields needed to calculate precipitation must be one file. This is a consequence of the basic way MEC works - it reads the grib files, does all the processing and then reads the next file. The current version now just copies the data that is really needed, reducing the amount of data considerably - in the first version all inference output was copied.

If we want to save disk space we simply could remove the mec directory. This is what could be done once this workflow is consolidated. Then no disk space is used unnecessarily at the end of the workflow. The feedback files are stored separately.

If the grib writing will be in one file - that would solve this as well.

I added a docstring explaining what this rule does.

workflow/rules/inference.smk
andreaspauling: I could use Snakemake's touch() on line 199 in inference.smk and then remove those three lines from the script in each function - would that adress your point? I could do that and test it.

andreaspauling: I tried to use touch(.../ok-file) in inference.smk instead of touching it in inference_prepare.py. I found no solution that worked. May we leave it with the current working solution or have a look at it together?

config/forecasters-ich1_mec_ffv2.yaml
jonasbhend, line 1: Any chance we can integrate the additional mec config in the existing templates instead of having to support an additional template? Or will this break the existing templates when run without the mec / ffv2 flags?

resources/ffv2/template_SYNOP_DET.nl.jinja2
```diff
@@ -0,0 +1,84 @@
+
+# variables to verify.
+varnoContinuous     'T2M,TD2M,RH2M,U10M,V10M,PS,FF,DD,GUST_6h,RR_6h' #,RR_6h,N,N_L,N_M,N_H,RAD_GL_1h
```
jonasbhend: Variables are hardcoded, does this work also with a subset? Should/can this be made configurable?

resources/ffv2/template_SYNOP_DET.nl.jinja2
```diff
@@ -0,0 +1,84 @@
...
+pecthresholds       list('FF'=list('lower'=c(2),'upper'=c(7))) # hit rates (percent correct forecast) for the forecast to hit the observation within the given limits are calculated.
+catthresholds       list('T2M'=c(282,292),'FF'=c(2.5,5,10)) # no space allowed between variables
```
jonasbhend: Any chance we can reuse the thresholds defined in the main config?

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
jonasbhend: There is a global postprocessor defined that is applied to all outputs just above. So the specific postprocessors for accumulation of precipitation introduced here are redundant.

src/evalml/cli.py
```diff
@@ -151,7 +151,7 @@ def execute_workflow(
     if not verbose:
         command += ["--quiet", "rules"]  # reduce verobosity of snakemake output
 
-    raise SystemExit(run_command(command))
+    return run_command(command)
```
jonasbhend, line 154: Why is this change introduced?

src/evalml/cli.py
```diff
@@ -163,22 +163,54 @@ def cli():
...
+            verbose,
+            dry_run,
+            unlock,
+            report,
+            dag,
+            rulegraph,
+            extra_smk_args,
+        )
+        if rc:
+            raise SystemExit(rc)
```
jonasbhend, line 219: why the conditional?

src/evalml/cli.py
```diff
@@ -163,22 +163,54 @@ def cli():
...
+    raise SystemExit(0)
```
jonasbhend, line 220: This seems superfluous and could  be omitted.

src/evalml/cli.py
jonasbhend, line 1: `execute_workflow` is now wrapped in `raise SystemExit()`. Why is this necessary and wouldn't it be cleaner to do inside the function to reduce duplication.

workflow/rules/verif_obs.smk
```diff
@@ -0,0 +1,366 @@
...
+                source_init=$(date -u -d "@$((src_epoch - lead * 3600))" +"%Y%m%d%H%M")
+                src_rel="$source_init/grib/${{source_init}}_${{lead3}}.grib"
+
+                if [[ -e "$src_rel" ]]; then
+                    dest="mec/{wildcards.init_time}/input_mod/${{source_init}}.grib"
+                    if [[ "$lead" -eq 0 ]]; then
+                        echo "Copying $src_rel -> $dest"
+                        cp "$src_rel" "$dest"
+                    else
+                        prev_lead3=$(printf "%03d" "$((lead - 6))")
```
jonasbhend: Is this coming from TOT_PREC_6H or why are we hardcoding six hourly intervals?

src/evalml/cli.py
```diff
@@ -163,22 +163,54 @@ def cli():
...
-        dry_run,
-        unlock,
-        report,
-        dag,
-        rulegraph,
-        extra_smk_args,
-    )
+    targets = []
+    if mec or ffv2:
+        if mec:
```
jonasbhend, line 198: I suggest to check here that mec and ffv2 blocks are present in config and fail fast if not.

workflow/scripts/generate_mec_namelist.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def _parse_steps(steps: str) -> int:
+    # check that steps is in the format "start/stop/step"
+    if "/" not in steps:
+        raise ValueError(f"Expected steps in format 'start/stop/step', got '{steps}'")
+    if len(steps.split("/")) != 3:
+        raise ValueError(f"Expected steps in format 'start/stop/step', got '{steps}'")
+    start, end, step = map(int, steps.split("/"))
+    return list(range(start, end + 1, step))
```
jonasbhend, line 21: This is duplicated from verif_obs.smk. I suggest to avoid this duplication. The tricky bit is because it is a hidden function, so we wouldn't want to import from elsewhere I guess.

jonasbhend (commented):
Hi @andreaspauling. Thanks for putting this together, it has grown quite big indeed so I only managed to have a broad look. The thing that worries me the most, is that we seem to limit the mec ffv2 pipeline (inadvertently) to evaluations of consecutive initializations spaced at most 6 hours apart (or step hours, I am not quite sure). I.e. we can evaluate initializations every 6, 3, or 1h, but this is prohibitively expensive to do for a full year and we haven't been doing that in the past.  I guess this stems from the orthogonal approach in mec/ffv2 verification (based on valid_time) and current evalml (based on initialization time). So my question is, if it would at all be possible to reorganize mec / ffv2 evaluation to organize feedback files by initialization, which would align the different pipelines and be much more flexible for future use.

config/forecasters-ich1_mec_ffv2.yaml
andreaspauling: Integrating the mec/ffv2 part in to an existing config should be possible. It is a design choice I guess. I have a slight preference for a larger number but small configs but I could try integration. Which existing one would then be best suited?

resources/ffv2/template_SYNOP_DET.nl.jinja2
andreaspauling: It works also with a subset. Is there a use case for subsets? We may always want to verify all variables we have.
This list depends on the variables in MEC and thus the observation (and model) source. If we decide on the observation source (see https://meteoswiss.atlassian.net/wiki/spaces/MR/pages/1746993334/SDL-42+Observation+sources+for+MEC+in+evalml ) the variables should stay constant

resources/ffv2/template_SYNOP_DET.nl.jinja2
andreaspauling: Agree, this should go into the config because that will be frequently changed. I will do it.

src/evalml/cli.py
andreaspauling: Honestly, Claude made this change...:
The reason for the change was the --mec --ffv2 sequential execution requirement. With raise SystemExit(run_command(command)), the first call to execute_workflow would immediately exit the Python process — so the second target could never run.

Returning the exit code instead (return run_command(command)) lets the experiment command loop over targets, check the code, and only exit after all targets have run.

src/evalml/cli.py
andreaspauling: Because rc == 0 means success. If MEC succeeds, the loop must continue to run FFV2 — raising SystemExit(0) after the first target would exit the process before FFV2 ever starts.

The condition implements "fail fast": if MEC fails (rc != 0), propagate the error immediately and skip FFV2 (which would be pointless anyway since its input files weren't produced).

src/evalml/cli.py
andreaspauling: This apparently makes sense, See comment above

src/evalml/cli.py
andreaspauling: Claude: `raise SystemExit` is necessary because Click command functions must either return `None `or raise to set the process exit code — if `execute_workflow `just returned a non-zero integer without raising, the process would exit with code 0.

Moving it inside `execute_workflow `would break `experiment`. That command needs `execute_workflow `to return the code so it can inspect it, conditionally abort, and only exit after all targets have run. If `execute_workflow `raised internally, the loop would never reach the second target.

So the duplication is the direct consequence of having two different call patterns:

`showcase`/`sandbox`/`make`: run once, exit immediately with the code
`experiment`: run once or twice, exit only after all targets
One way to eliminate the duplication in the single-target commands while keeping `experiment `working:


```
def execute_workflow(...) -> int:
    return run_command(command)

def _run_workflow(...):
    raise SystemExit(execute_workflow(...))
```
Then `showcase`/`sandbox`/`make `call `_run_workflow`. Three `raise SystemExit(execute_workflow(...))` calls is not much duplication to be concerned about.

workflow/rules/verif_obs.smk
andreaspauling: Yes, this is because we use 6h precip at the moment. Should be made more flexible later

src/evalml/cli.py
andreaspauling: That sounds good to me. I will do it.

workflow/scripts/generate_mec_namelist.py
andreaspauling: True. However, deduplicating it is awkward (Claude):

- `generate_mec_namelist.py` is a standalone script run via `uv run` — it can only import from packages installed in the environment, not from `workflow/rules/`.

`verif_obs.smk` is loaded by Snakemake — it can't easily import from `workflow/scripts/`.

One option is to move `_parse_steps` into the `evalml `package (e.g., `src/evalml/utils.py`) and import it in both places. But that means the script would need evalml as a dependency (it's run with `uv run`, so it depends on what's in the project environment).
Given the duplication is just 5 lines, I'd leave it as-is.

jonasbhend: ok, i guess

workflow/rules/verif_obs.smk
jonasbhend: The problem being that evalml can evaluate n-hourly precip. It is not explicit in evalml so far because it depends on the step of the forecasts. My suggestion would be to add a comment in the code to highlight that this may need changing.

andreaspauling:
> Hi @andreaspauling. Thanks for putting this together, it has grown quite big indeed so I only managed to have a broad look. The thing that worries me the most, is that we seem to limit the mec ffv2 pipeline (inadvertently) to evaluations of consecutive initializations spaced at most 6 hours apart (or step hours, I am not quite sure). I.e. we can evaluate initializations every 6, 3, or 1h, but this is prohibitively expensive to do for a full year and we haven't been doing that in the past. I guess this stems from the orthogonal approach in mec/ffv2 verification (based on valid_time) and current evalml (based on initialization time). So my question is, if it would at all be possible to reorganize mec / ffv2 evaluation to organize feedback files by initialization, which would align the different pipelines and be much more flexible for future use.

Thanks @jonasbhend for the review! There is no limitation from mec/ffv2 regarding the initialization spacing. You may be referring to the hardcoded 6h, which are steps. This is because we used TOT_PREC_6h so far. Absolutely agree that this needs more flexibility. This can also be done without reorganizing the whole mec /ffv2 evaluation because it is well confined in the set up of the mec cases. However, I would do that in a new PR because this one is already too heavy...

resources/ffv2/template_SYNOP_DET.nl.jinja2
jonasbhend: Well, as you say this depends on the observations AND model data. So far we don't have wind gusts for example, so this shouldn't work. If there is internal homogenization (subsettting) of the requested parameters to the intersect of params available from both forecast and ground truth, this won't break with an 'extensive' set, but I still think it would be clearer if specified accordingly.

workflow/rules/verif_obs.smk
andreaspauling: Good idea. I will do it.
I also create a Jira task, so that it gets done more quickly

jonasbhend:
> > Hi @andreaspauling. Thanks for putting this together, it has grown quite big indeed so I only managed to have a broad look. The thing that worries me the most, is that we seem to limit the mec ffv2 pipeline (inadvertently) to evaluations of consecutive initializations spaced at most 6 hours apart (or step hours, I am not quite sure). I.e. we can evaluate initializations every 6, 3, or 1h, but this is prohibitively expensive to do for a full year and we haven't been doing that in the past. I guess this stems from the orthogonal approach in mec/ffv2 verification (based on valid_time) and current evalml (based on initialization time). So my question is, if it would at all be possible to reorganize mec / ffv2 evaluation to organize feedback files by initialization, which would align the different pipelines and be much more flexible for future use.
> 
> Thanks @jonasbhend for the review! There is no limitation from mec/ffv2 regarding the initialization spacing. You may be referring to the hardcoded 6h, which are steps. This is because we used TOT_PREC_6h so far. Absolutely agree that this needs more flexibility. This can also be done without reorganizing the whole mec /ffv2 evaluation because it is well confined in the set up of the mec cases. However, I would do that in a new PR because this one is already too heavy...

@andreaspauling I don't think this is the case. Due to `_reftimes_mec` this only works for initializations at least every `step` (here 6) hours.

The following is the cropped output from running evalml (dry mode) with initializations every 6h, i.e. the follwoing in `forecasters-ich1_mec_ffv2.yaml`:
```
dates:
  start: 2025-07-01T00:00
  end: 2025-07-30T00:00
  frequency: 6h
```
the list of tasks is as expected:
```
> evalml experiment config/forecasters-ich1_mec_ffv2.yaml -n --ffv2
host: balfrin-ln002
Building DAG of jobs...
Job stats:
job                               count
------------------------------  -------
ffv2_all                              1
generate_ffv2_namelist                1
generate_mec_namelist               113
inference_create_venv                 1
inference_execute                   117
inference_extract_requirements        1
inference_get_checkpoint              1
inference_make_squashfs_image         1
inference_prepare_forecaster        117
link_mec_input                      113
prepare_mec_input                   113
reorganize_ffv2_files                 1
run_ffv2                              1
run_mec                             113
sarus_pull_ffv2                       1
sarus_pull_mec                        1
total                               696
```
if I run with initializations every few days
```
dates:
  start: 2025-07-01T00:00
  end: 2025-07-30T00:00
  frequency: 30h
``` 
I get 
```
> evalml experiment config/forecasters-ich1_mec_ffv2.yaml -n --ffv2
host: balfrin-ln002
Building DAG of jobs...
Job stats:
job                       count
----------------------  -------
ffv2_all                      1
generate_ffv2_namelist        1
reorganize_ffv2_files         1
run_ffv2                      1
sarus_pull_ffv2               1
total                         5
```
whereas the task list is the expected when running without the mec/ffv2 flag:
```
> evalml experiment config/forecasters-ich1_mec_ffv2.yaml -n
host: balfrin-ln002
Building DAG of jobs...
Job stats:
job                                 count
--------------------------------  -------
experiment_all                          1
inference_create_venv                   1
inference_execute                      24
inference_extract_requirements          1
inference_get_checkpoint                1
inference_make_squashfs_image           1
inference_prepare_forecaster           24
report_experiment_dashboard             1
verification_metrics                   24
verification_metrics_aggregation        1
verification_metrics_plot               1
write_summary                           1
total                                  81
```

config/forecasters-ich1_mec_ffv2.yaml
jonasbhend: we are supporting three forecaster configs (I don't know why we even have these 3 and they seem all a bit outdated). So I would have suggested to integrate it with one (or all) of these. But honestly, now that I look at the config mess, I am not even sure this is the way to go...

config/forecasters-ich1_mec_ffv2.yaml
```diff
@@ -0,0 +1,93 @@
...
+  start: 2025-07-26T00:00
+  end: 2025-07-27T00:00
+  frequency: 6h
+
+runs:
+  - forecaster:
+      inference_resources:
+          slurm_partition: normal-shared
+      checkpoint: https://service.meteoswiss.ch/mlstore#/experiments/602/runs/c30490b6ba064e4db03b430f3a2595ad
+      label: stage_E_icon_1km_cutoff_edges_subgrid_horography
```
jonasbhend:
```suggestion
      label: stage_E_icon_1km_cutoff_edges_subgrid_orography
```

resources/ffv2/template_SYNOP_DET.nl.jinja2
andreaspauling: FFV2 just ignores a variable if it is not there. But I agree it would be clearer to specify only those variables we actually have.

resources/ffv2/template_SYNOP_DET.nl.jinja2
jonasbhend: So we could basically specify the most extensive list of parameters availbale from observations and then be done with it? Is this the case now?

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
andreaspauling: Do you mean
          - accumulate_from_start_of_forecast: # accumulate tp from start of forecast
              accumulations:
                - tp
can be removed?

resources/ffv2/template_SYNOP_DET.nl.jinja2
andreaspauling: The variables are reduced to the ones we have

resources/ffv2/template_SYNOP_DET.nl.jinja2
andreaspauling: The thresholds are now specified in the config. See forecasters-ich1_mec_ffv2.yaml

workflow/rules/verif_obs.smk
andreaspauling: Done.

src/evalml/cli.py
andreaspauling: Done

config/forecasters-ich1_mec_ffv2.yaml
andreaspauling: You were right regarding the initializations being max. 6 hours apart. That should be fixed now, thanks for pointing at this.

andreaspauling:
Main is merged and all necessary folow-up changes in the MEC/FFV2 parts done.

config/forecasters-ich1_mec_ffv2.yaml
```diff
@@ -0,0 +1,124 @@
...
+  catthresholds:
+    T2M: [282.0, 292.0]
+    TD2M: [260.0, 273.0]
+    RH2M: [0.9, 1.0]
+    U10M: [5.0, 10.0]
+    V10M: [5.0, 10.0]
+    PS: [100000.0, 101000.0]
+    RR_6h: [5.0, 10.0]
+  pecthresholds:
+    T2M:
```
jonasbhend, line 91: Thresholds for binary categorical forecasts are already defined in the experiments section. Is there a particular reason to redefine them in two places in the config? What is the difference between pecthresholds and catthresholds?

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
jonasbhend: Yes, the one you introduced for grib files only in lines 36-38 is redundant and can be removed.

jonasbhend (approved):
Thanks for the perseverance.
I have only taken a broad look and have given it a spin with the example config. Works fine for me.

config/forecasters-ich1_mec_ffv2.yaml
andreaspauling: Good point. We should check if it makes sense and how it can be done to satisfy both purposes. I made a Jira task for it.
catthresholds are used for categorical verification and pectthresholds are used for hit rates (percent correct forecast) for the forecast to hit the observation within the given limits.

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
andreaspauling: Done in a new commit

frazane (approved):
Good to go, thanks for the patience @andreaspauling!

PR #125, opened 2026-04-01
Guidelines for workflow definition

frazane:
This PR introduces guidelines for extending the evalml workflow and enforces them. It's almost entirely renaming some rules and files so it should not break anything, but I would test anyway.

* Rename verif.smk → verification.smk and all verif_* rules to verification_* to match the {module}_ prefix convention
* Rename rules to follow {module}_{operation} pattern (e.g. prepare_checkpoint → inference_get_checkpoint, create_inference_venv → inference_create_venv)
* Rename scripts to mirror their calling rules (e.g. extract_baseline.py → data_extract_baseline.py, verif_single_init.py → verification_metrics.py)
* Remove vzarr_references.py (dead code)
* Add a "Workflow development guidelines" section to the README documenting rule/script naming, log paths, output layout, and wildcard conventions


Louis-Frey (approved):
So I tested all the configs, with the exception of `forecasters-co1e.yaml` (this was expected to fail), and before main was merged into this branch. 

`forecasters-ich1-oper-fixed.yaml` produces some errors when initially run on the 2024 cases, but when I run it first on the 2025 case, and then on the 2024 cases again, it runs fine. I will take it low-prio on my list to look into that as well. 

The other configs run without error. Good to go from my side.

frazane:
> forecasters-ich1-oper-fixed.yaml produces some errors when initially run on the 2024 cases, but when I run it first on the 2025 case, and then on the 2024 cases again, it runs fine. I will take it low-prio on my list to look into that as well.

Since this is the config that we use for the latest experiments, I would not merge this until this is fixed. What was the error?

README.md
```diff
@@ -149,3 +149,87 @@ ln -s $SCRATCH/evalenv/output output
...
+```
+
+Multiple wildcards are joined with a hyphen: `{run_id}-{init_time}.log`.
+
+### Output directory layout
+
+All outputs are rooted at `OUT_ROOT` (from `locations.output_root` in the config):
+
+```
+{OUT_ROOT}/
```
dnerini, line 205: I think this bit needs to be updated after #122 ?

README.md
```diff
@@ -149,3 +149,87 @@ ln -s $SCRATCH/evalenv/output output
...
+└── results/{experiment_name}/               # final products
+    ├── dashboard/
+    └── plots/
+```
+
+### Wildcard conventions
+
+| Wildcard | Format | Example |
+|---|---|---|
+| `{run_id}` | `{type}-{ckpt_hash}-{run_hash}` | `forecaster-1a2b-c3d4` |
```
dnerini: also need to be update following #122

dnerini (commented):
thanks @frazane for this very much needed cleanup and documentation effort!

looks very good to me, but before merging it should be updated following the changes introduced by #122

README.md
dnerini: see ee0757f

README.md
dnerini: see ee0757f

dnerini (approved):

PR #130, opened 2026-04-07
Mrb 807 avoid regridding forecasts

Louis-Frey:
Avoid unnecessary reprojection / regridding if forecasts and obs are on the same grid.

1. Performance of `map_forecast_to_truth` is as follows (full forecast, all parameters):
    a. With early exit (i.e. just copy coordinates): 0.02s
    b. With full reprojection: 1.5s
2. Differences between KENDA-CH1 analysis latitude / longitude and ICON-CH1-EPS baseline or Varda grids can be much larger than 'obvious' rounding errors. A quick test revealed, that these are up to `max lon diff 0.0002441` which roughly corresponds to 20 m. The current implementation sets the tolerance to approx. 30 m (`0.0003`) in both lat and lon, so that these avoid full reprojection. 

This performance gain doesn't matter for experiment runs (standard scores) but may be considerable for the computation of maps, where `map_forecast_to_truth` is invoked many times.

### Summary of changes
* check (with tolerance) if fcst and obs are on the same grid
* if they are, copy coordinates from truth (to avoid rounding errors) and return fcst directly


tests/unit/test_spatial_mapping.py
```diff
@@ -91,6 +91,37 @@ def test_map_forecast_to_truth_maps_forecast_to_truth_locations():
...
+            "y": [0, 1],
+            "x": [0, 1],
+            "lat": (("y", "x"), lat),
+            "lon": (("y", "x"), lon),
+        },
+    )
+
+    result = map_forecast_to_truth(fcst, truth)
+
+    assert result is fcst
```
jonasbhend, line 123: Hi @Louis-Frey. Great addition. The below is just a suggestion:

This doesn't really test if we abort early, no? Either the test should be expanded to actually test timing (probably difficult), or one could add a log statement to the if branch to check if the early option is taken as suspected (and check if not if the grids differ).

Louis-Frey: Hi Jonas, thanks! I will have a look...

jonasbhend:
Ok, @frazane @Louis-Frey @dnerini I have expanded the tests slightly to make sure we can align forecasts even when skipping interpolation. Can someone of you have a close look at the tests to make sure we have caught all the edge cases.

jonasbhend:
@clairemerker if you have time to look into this, your feedback would be very much appreciated. For context, I worry about the edge cases that render alignment of forecast and ground truth with xarray impossible:
1. ICON grid with rounding errors in lon/lat 
2. ICON grid with corresponding lon/lat coordinates but different 'values' dimension indices

I don't know if the latter could occur. Clearly if stored as anything other than int, the dimension values could also be affected by rounding errors, which I don't think is a thing, but anyway...

dnerini:
> I worry about the edge cases 

but we essentially work with always the same 2 or 3 models grids, can't we just test each of them to make sure it works?
and in the worst case, what will happen is that we'll be regridding some data needlessly, which is not such a big deal, no?

jonasbhend:
> > I worry about the edge cases
> 
> but we essentially work with always the same 2 or 3 models grids, can't we just test each of them to make sure it works? and in the worst case, what will happen is that we'll be regridding some data needlessly, which is not such a big deal, no?

Agreed, I think the key point is that skipping the interpolation when coordinates are the same withing rounding errors is unsafe, as this potentially may result in empty xarray joins after alignment. The current implementation tests for that so we should be good to go. I don't think we need to test the different grids explicitly, as this would complicate the unit tests or bloat the repos (we'd have to check in the grid specifications, right)?

PR #131, opened 2026-04-07
Fix erroneous TOT_PREC loading for runs and baselines

Louis-Frey:
## Summary

Fix precipitation handling in evalml so that TOT_PREC is read correctly for both runs and baselines, and add defensive checks against silent corruption.

### Inference configs (`resources/inference/configs/`)
- Add the `accumulate_from_start_of_forecast` post-processor to all forecaster GRIB outputs (LAM and IFS) and to the interpolator configs. anemoi-inference now writes TOT_PREC as cumulative-from-start, matching the operational baseline-zarr convention.

### Data loading (`src/data_input/__init__.py`)

`load_baseline_from_zarr`:
- Restrict to the requested lead times **before** disaggregating, so that `.diff()` operates on the requested step interval (e.g. 6 h) rather than the zarr's hourly steps. This fixes the original "1 h precip reported as 6 h precip" bug.

`load_fct_data_from_grib`:
- Disaggregate cumulative-from-start TOT_PREC with `.diff()` (was removed temporarily in this branch when the GRIB was period-accumulated; reinstated now that the post-processor produces cumulative output).
- Force TOT_PREC at lead_time = 0 to 0 before `.diff()`. anemoi-inference omits step 0 of TOT_PREC from the GRIB even with the post-processor enabled; the resulting NaN would otherwise propagate through `.diff()` and wipe out the first period accumulation.

Both loaders:
- Replace fragile `.pad(constant_value=None)` with explicit `.reindex(lead_time=lead_times)` so that lead_time = 0 carries a real coordinate value (NaT-vs-0 alignment was relying on xarray's implicit behaviour).
- Sanity-check that the input is actually cumulative: if `min(.diff())` is below −0.1 mm, raise `ValueError` rather than silently producing garbage. Float-noise negatives are clipped to 0.

## Test plan

- [x] Short benchmark (2 weeks of 6-hourly inits, March 2025) with `stage_E_realch1` ML forecaster, ICON-CH1-EPS and ICON-CH2-EPS baselines, KENDA-CH1 truth.
- [x] Resulting `RMSE_TOT_PREC.png`: no values at lead_time = 0, near-monotonic RMSE growth for all three sources, expected ordering (forecaster < ICON-CH1 < ICON-CH2).
- [x] test all ICON-emulator example configs.

Co-authored-by: Francesco Zanetta <francesco.zanetta@meteoswiss.ch>


frazane (approved):
Looks good! Could you post a screenshot of a dashboard with the updated results, for future reference?

Louis-Frey:
> Could you post a screenshot of a dashboard with the updated results

Yes, good point. Stay tuned.

Louis-Frey:
Turns out the fix did not solve the problem. I'm investigating and will keep you posted.

Louis-Frey:
Ok, the issue is clear now. On the one hand, there was a bug, computing diffs of 6-hourly precipitation sums that was intended for accumulations since the start of the forecast. This is what the PR fixes. 

Then, the lead time plots with RMSE averaged over forecasts still did not look as one would expect, but this is just because individual forecasts have huge variability that is not averaged out over a small sample. 

More information can be found in the corresponding Jira Task: https://meteoswiss.atlassian.net/browse/MRB-808?atlOrigin=eyJpIjoiNjdjNzRhZDU0OGY5NGVlZjhhNjkwOTkwOTMxZmE1OTYiLCJwIjoiaiJ9 (ok to post that here?)

I will proceed to testing the configs now.

Louis-Frey:
Ok, I tested all the ICON configs, and the `forecasters-co2.yaml` config, which all ran smoothly, with the exception of the 2024 issue in `forecasters-ich1-oper-fixed.yaml` which is ok. So from my side, the PR is ready to merge.

Please check again @dnerini @frazane

src/data_input/__init__.py
```diff
@@ -117,15 +117,18 @@ def load_fct_data_from_grib(
...
+        # already contains the precip for that period (startStep = endStep - period),
+        # which is directly comparable to the analysis zarr's TOT_PREC_6H.
+        # Do NOT apply diff here — that would only be correct for cumulative-from-start
+        # data (as in the operational baseline zarr handled by load_baseline_from_zarr).
+        # Do NOT apply fillna(0) — step 0 has no TOT_PREC in the GRIB (the encoding
+        # shifts the date for negative startStep), so the structural NaN must be
+        # preserved to avoid fake zero-precipitation at lead time 0.
+        # Also, do not clip negative precipitations to zero. Seeing negative precipitations
+        # when they are there is an important sanity-check, e.g. in Meteograms.
+        LOG.info("Precipitation is already period-accumulated, no further processing")
```
jonasbhend: Do we intend to check if precipitation is period-accumulated in the future. If not, I suggest to get rid of the comments and empty if clause.

src/data_input/__init__.py
```diff
@@ -151,22 +154,23 @@ def load_baseline_from_zarr(
...
-                .pad(lead_time=(1, 0), constant_value=None)
-                .clip(min=0.0)
-            )
-        )
+        ## disaggregate precipitation: select the requested lead times BEFORE
+        ## differencing so that .diff() computes accumulations over the full
+        ## step interval (e.g. 6 h) rather than between consecutive hourly
+        ## steps in the zarr.
+        ## Do NOT apply fillna(0) — if a selected step is genuinely missing,
+        ## it is better to see NaN than to silently substitute zero precip.
```
jonasbhend: Filling with NA was done based on the following observations / rationale: 
In COSMO / ICON baselines, TOT_PREC at lead_time 0 was missing (not in grib file), therefore, precipitation amounts for the first 6/1 hour(s) couldn't be computed. `fillna(0)` fixes this (with the noted side-effect of masking any 'real' missing values.

src/data_input/__init__.py
```diff
@@ -151,22 +154,23 @@ def load_baseline_from_zarr(
...
+        cumul = baseline.TOT_PREC.sel(lead_time=lead_times)
+        precip = cumul.diff("lead_time").reindex(lead_time=lead_times)
```
jonasbhend: `cumul.diff("lead_time")` results in one fewer lead times than selected. In my view, this generates excessive missing values (at beginning / or end). This is ok if steps include 0, as we cannot compute `TOT_PREC` at time zero, but if steps were to start at some time > 0, then we lose potentially available information.

src/data_input/__init__.py
Louis-Frey: Thanks for checking Jonas. I should have added you too.

No, I did not plan on implementing such a test. For the moment, I would just stick to the convention that all precipitation in EvalML is period-accumulated. If ever anemoi-inference changes back again to accumulations since the start of the forecast, it will be easier to spot this time, since precipitation amounts will be monotonically increasing (no diff being done). This time it was hard to spot since the negative precip values resulting from the diff were set to zero.

So I would leave it for the moment, but yes we can get rid of the comments and the if-clause.

src/data_input/__init__.py
Louis-Frey: But in this case it's a real missing value anyway since we don't have this information. Why would we want to fill that up with zero?

As I understood, the missing value for lead time zero results from the fact that it would be an accumulation since before the forecast started (e.g., lead time -6h and 0h). In this case it would make sense to keep the NaN because we simply don't have that information.

src/data_input/__init__.py
Louis-Frey: Yes, that's true. I could build a workaround that will handle this edge case (i.e., also load the previous step if the first step is >0h), but I wonder if it will be worth it. I don't suppose we will have that case very often, if ever. (?)

src/data_input/__init__.py
```diff
@@ -156,22 +156,28 @@ def load_baseline_from_zarr(
...
                 .diff("lead_time")
                 .pad(lead_time=(1, 0), constant_value=None)
                 .clip(min=0.0)
             )
         )
     baseline = baseline[params].sel(
         ref_time=reftime,
-        lead_time=np.array(steps, dtype="timedelta64[h]"),
+        lead_time=lead_times,
     )
```
jonasbhend: We could simply do the sel statement before going in the TOT_PREC if clause, that way we already have the corresponding lead times selected and don't have issues with 6h/1h precipitation.

jonasbhend:
@Louis-Frey if you still want to push this, the only thing that really needs fixing is the lead time selection before disaggregation to allow 6h precip to be evaluated (but this will be tackled in a separate PR to make aggregation explicity in evalml anyway, so is obsolete). As a courtesy, one could check if aggregated precip is strictly increasing (to within rounding errors) and warn/fail if this is not the case, but this is rather a nice to have.

dnerini:
also wonder if we should merge this with #135? thoughts?

workflow/rules/verification.smk
dnerini, line 1: I suppose this change should be reverted?

Louis-Frey: Yes, I will still do so and give you a heads-up once I think everything is ready.

Louis-Frey:
> also wonder if we should merge this with #135? thoughts?

Claude's take:

```
don't merge them. Reasons:

1. Different scopes — easier to review and revert each independently if something goes wrong.
2. Different base branches — combining would require sorting out which base to target.
3. This PR has been through a lot of iteration already; adding more changes risks delaying it further.
4. If #135 needs to be rebased onto main eventually anyway, that's a separate cleanup.
```

Sounds reasonable in my view, particularly 2. and 3. The partial redundancies in adding the `accumulate_from_start_of_forecast` post-processor to forecasters / interpolators can be easily reconciled.

Louis-Frey:
Testing the ICON example configs now. Will give you a heads-up once ready.

Louis-Frey:
After pinning the anemoi-inference version to 0.10.0, all ICON example configs have run without error. The PR is ready. I also updated the PR description. Please check again @dnerini @jonasbhend @frazane

jonasbhend (approved):
Thanks, very thorough

PR #132, opened 2026-04-20
Fix wrong wildcard used in inference rules

frazane:
This pull request is a hotfix that updates some rules in `inference.smk` that were using the wrong wildcard. The wrong wildcards were introduced in #125.


dnerini (approved):
LGTM!

PR #135, opened 2026-04-23
Add new colormaps for precip 1h and 6h accumulations

dnerini:


clairemerker (approved):
Looks good, thanks! :)

workflow/rules/plot.smk
```diff
@@ -100,16 +100,19 @@ rule plot_forecast_frame:
         grib_out_dir=lambda wc: (
             Path(OUT_ROOT) / f"data/runs/{wc.run_id}/{wc.init_time}/grib"
         ).resolve(),
+        accu=lambda wc: int(RUN_CONFIGS[wc.run_id]["steps"].split("/")[2]),
```
jonasbhend, line 103: using a `np.diff(steps)[0]` approach would be a bit more robust, in case someone runs an evaluation without the analysis.

dnerini: good point, so you would implement that directly in the plotting script instead of passing `accu` as a parameter?

Louis-Frey (approved):
Thanks Daniele, looks good from my side!

PR #136, opened 2026-04-24
Keep _is_candidate=True if the run was already registered

dnerini:
`runs |= register_run(...)` would silently overwrite any existing entry, including the `_is_candidate` flag. When a forecaster appeared before the interpolator that depends on it, the interpolator's recursive registration (with `as_candidate=False`) would clobber the forecaster's already-set True.

The fix iterates the new entries individually and keeps `_is_candidate=True` if the run was already registered as an explicit candidate, regardless of config order.


frazane (approved):
Looks good, thanks!

PR #137, opened 2026-04-27
Categorical scores

jonasbhend:
### Overview of Changes
* configure thresholds for categorical scores by parameter in main configs (e.g. interpolators-ich1.yaml)
* compute and store contingency matrix (i.e. counts of true positives, ...) per initialization in `verif.nc`
  * threshold.<param> is a new dimension that is being introduced to efficiently store the contingency matrix (without having to store the threshold value in the variable name).
* compute aggregated contingency matrix and categorical scores in `verification_aggregation.py`
  * generate a `scores.BasicContingencyManager` object from aggregated counts
  * compute categorical metrics from contingency manager object 
* _added a new function to decode sanitized metric strings (e.g. `ETS_gt_0p0` to `ETS > 0.0`)_  

The categorical scores are selected by name in the dashboards, as thresholds do not align. This leads to many new variables being available for selection in the dashboard with lots of those selections being empty  (for variables that have not been evaluated using the particular threshold). If you have a better suggestion to solve this, please let me know.

### Example dashboard 
live version here ["M:\zue-prod\fc_development\seamless\S-RUC\evaluation\MRB-818_dashboard.html"](file:///M:/zue-prod/fc_development/seamless/S-RUC/evaluation/MRB-818_dashboard.html)

image

### Naming of variables and schema of verification netcdfs

1. single-initialization file
```
<xarray.Dataset> Size: 227kB
Dimensions:                     (ref_time: 1, region: 1, source: 2,
                                 lead_time: 121, T_2M.threshold: 3,
                                 contingency: 5, U_10M.threshold: 3,
                                 V_10M.threshold: 3, TOT_PREC.threshold: 3)
Coordinates:
  * region                      (region) <U3 12B 'all'
  * source                      (source) <U13 104B 'ICON-CH2-ctrl' 'KENDA-CH1'
  * lead_time                   (lead_time) timedelta64[ns] 968B 00:00:00 ......
  * contingency                 (contingency) <U11 220B 'tp_count' ... 'total...
  * T_2M.threshold              (T_2M.threshold) <U9 108B 'lt_273p15' ... 'gt...
  * ref_time                    (ref_time) datetime64[ns] 8B 2025-03-01
    time                        (lead_time) datetime64[ns] 968B ...
  * U_10M.threshold             (U_10M.threshold) <U7 84B 'gt_2p5' ... 'gt_10p0'
  * V_10M.threshold             (V_10M.threshold) <U7 84B 'gt_2p5' ... 'gt_10p0'
  * TOT_PREC.threshold          (TOT_PREC.threshold) <U8 96B 'gt_0p0' ... 'gt...
Data variables: (12/60)
    T_2M.BIAS                   (ref_time, region, source, lead_time) float64 2kB ...
    T_2M.MSE                    (ref_time, region, source, lead_time) float64 2kB ...
    T_2M.MAE                    (ref_time, region, source, lead_time) float64 2kB ...
    T_2M.CORR                   (ref_time, region, source, lead_time) float64 2kB ...
    T_2M.contingency_table      (ref_time, region, source, T_2M.threshold, contingency, lead_time) float64 29kB ...
...
```
2. aggregated verification file
```
<xarray.Dataset> Size: 916kB
Dimensions:                (season: 2, init_hour: 2, region: 1, source: 2,
                            lead_time: 121, contingency: 5, T_2M.threshold: 3,
                            U_10M.threshold: 3, V_10M.threshold: 3,
                            TOT_PREC.threshold: 3)
Coordinates:
  * region                 (region) <U3 12B 'all'
  * source                 (source) <U13 104B 'ICON-CH2-ctrl' 'KENDA-CH1'
  * lead_time              (lead_time) timedelta64[ns] 968B 00:00:00 ... 5 da...
  * init_hour              (init_hour) int64 16B -999 0
  * season                 (season) <U3 24B 'MAM' 'all'
  * contingency            (contingency) <U11 220B 'tp_count' ... 'total_count'
  * T_2M.threshold         (T_2M.threshold) <U9 108B 'lt_273p15' ... 'gt_298p15'
  * U_10M.threshold        (U_10M.threshold) <U7 84B 'gt_2p5' 'gt_5p0' 'gt_10p0'
  * V_10M.threshold        (V_10M.threshold) <U7 84B 'gt_2p5' 'gt_5p0' 'gt_10p0'
  * TOT_PREC.threshold     (TOT_PREC.threshold) <U8 96B 'gt_0p0' ... 'gt_0p005'
Data variables: (12/118)
    TOT_PREC.FAR_gt_0p0    (season, init_hour, region, source, lead_time) float64 8kB ...
    TOT_PREC.FAR_gt_0p001  (season, init_hour, region, source, lead_time) float64 8kB ...
    TOT_PREC.FAR_gt_0p005  (season, init_hour, region, source, lead_time) float64 8kB ...
    TOT_PREC.POD_gt_0p0    (season, init_hour, region, source, lead_time) float64 8kB ...
    ...
```


config/forecasters-co1e.yaml
```diff
@@ -39,6 +39,24 @@ stratification:
...
+    - 5.0
+    - 10.0
+  V_10M:
+    - 2.5
+    - 5.0
+    - 10.0
+  T_2M:
+    - 273.15 #  0.0 deg C
+    - 288.15 # 15.0 deg C
+    - 298.15 # 25.0 deg C
```
dnerini: don't we want to be able to specify the direction of the threshold (lower than, greater than), particularly for temperature?

workflow/scripts/verification_metrics.py
```diff
@@ -133,6 +141,12 @@ def main(args: ScriptConfig):
         help="Comma-separated list of shapefile paths defining regions for stratification.",
         default="",
     )
+    parser.add_argument(
+        "--threshold_dict",
+        type=lambda x: eval(x),
```
frazane, line 146: Wow, I didn't know this was an option. Very good to know!

config/forecasters-co1e.yaml
```diff
@@ -39,6 +39,24 @@ stratification:
     - alpensuedseite
   root: /scratch/mch/bhendj/regions/Prognoseregionen_LV95_20220517
 
+thresholds:
```
dnerini, line 42: I like the approach being taken by @cosunae in #139 , grouping all plotting options under a showcase config section. I wonder if we could do the same for the experiment options, something like:

```
experiment:
  dashboard: true
  scorecards: false
  thresholds:
    TOT_PREC:
      ...
```

dnerini (commented):
very cool addition, thanks! I left a couple of initial high-level comments, but I look forward to test this in more detail!

config/forecasters-co1e.yaml
dnerini: with this we could also simplify the evalml cli with just one `evalml run` command .... @frazane

jonasbhend: Should we push this all the way and separate out the experiment and showcase specific configuration into separate files, so we don't have to modify in multiple places?

workflow/scripts/verification_aggregation.py
frazane, line 1: Do you have a feeling of how good we are approximating the real scores by computing them per initialization time?

workflow/scripts/verification_aggregation.py
```diff
@@ -54,10 +63,55 @@ def aggregate_results(ds: xr.Dataset) -> xr.Dataset:
...
+                out[var.replace("contingency_table", metric)] = fn(contingency_manager)
+            # split by threshold dimension into data variables
+
+            out = out.drop_vars(var)
+
+    # second loop to split along thresholds dimension, could be moved to plotting/dashboards scripts
+    for var in out.data_vars:
+        dim = list(filter(lambda x: "threshold" in x, out[var].dims))
+        if dim:
+            rename_dict = {d: f"{var}_{d}" for d in out.coords[dim[0]].values}
```
frazane, line 89: Could it make sense to replace the decimal point with "p" to distinguish it from the point separating the variable name and the metric? (Really just a detail!)

config/forecasters-co1e.yaml
jonasbhend: Yeah, that was too hard ;-). so for now it is always greater than. I guess I could easily extend this to:
```
T_2M:
  - < 273.15
  - >= 288.15
```
and then match this to the operator down the line.

resources/report/dashboard/template.html.jinja2
frazane, line 1: Regarding the dashboard issue, I think it should be possible to filter out the variables that don't have certain metrics. It can be done with the JavaScript code in the dashboard. I'd try asking nicely to our dear friend.

src/verification/__init__.py
frazane, line 1: Could we separate the logic of the categorical scores in a separate function? Also, would be nice to add some comments especially on the `scores` functionality to help with those who are unfamiliar with the library.

resources/report/dashboard/template.html.jinja2
jonasbhend: @frazane I don't understand your comment. I really don't want to go to dynamic drop-downs, that is basically dashboard hell. The fundamental problem is the grid layout. There is no easy solution to condense the variable x metric (plus threshold) sparse matrix, as metrix can be available for multiple thresholds, else we could probably just put all ETS (if available and selected) on one line, but there may be multiple ETS scores for some of the variables and none for some others...

frazane: What I mean is that the dashboard automatically removes variables from the selection if a selected metric is not available for them.

frazane: If you prefer to have different layouts depending on the metric types, I have nothing against that.

jonasbhend: > What I mean is that the dashboard automatically removes variables from the selection if a selected metric is not available for them.

Yes, I guess this we could do. Currently, I prefer to just leave it to the user to clean up the sparse matrix if necessary, that way you have more flexibility in what you can display at once.

config/forecasters-co1e.yaml
jonasbhend: @daniele @frazane ?

src/verification/__init__.py
jonasbhend: Is this what you were asking for?

config/forecasters-co1e.yaml
jonasbhend: @dnerini @frazane I like the idea about the configs, but I also think we should do it outside of this PR. having said that, those quality-of-life improvements get higher and higher priority the more we delay these.

frazane: I agree on doing it in a separate PR. But let's meet to agree on a new improved interface rather soon!

src/evalml/config.py
```diff
@@ -227,6 +227,19 @@ class Stratification(BaseModel):
...
 
+class Threshold(BaseModel):
+    """Configuration for thresholds used in verification metrics."""
+
+    thresholds: Dict[str, List[str]] = Field(
+        default_factory=dict,
+        description=(
+            "Dictionary mapping parameter names to lists of thresholds. "
+            "These thresholds will be used for computing classification metrics at different cutoffs. "
+            "Example: {'temperature': [0, 10, 20], 'precipitation': [0, 1.0]}"
```
frazane: Isn't the example wrong?

src/verification/__init__.py
```diff
@@ -73,31 +84,92 @@ def _mask_from_polygons(
         return xr.DataArray(mask, coords=lon.coords, dims=lon.dims)
 
 
+# deconstruct the threshold string into the value and the operator
+
+
+def _threshold_value_and_operator(threshold: str):
+    """
+    Parse a threshold string like '> 10.0' into (operator.gt, 10.0).
+    Supported operators: >, >=, <, <=, ==, !=
```
frazane: I must say, I am not a big fan of using operators in the string like this. Don't we only need "higher than" and "lower than" anyway? I'd be more in favor of something like

```yaml
T_2M:
  low: [273.15]
  high: [288.15, 298.15]
``` 

or similar. It makes it also easier to validate the config based on the schema.

src/evalml/config.py
jonasbhend: yep, good catch.

src/verification/__init__.py
jonasbhend: Agreed, this is a bit hacky. But if we go for your proposition, I would rather do
```
T_2M:
   gt: [273.15]
   lt: [288.5, 298.15]
```
which allows us to use all the available operators without much configuration. Also, this could probably be used to sanitize the variable names in the aggregated verification files from e.g. `TOT_PREC.ETS > 0.0` to something like `TOT_PREC.ETS_gt_0p0` to avoid special characters. On the other hand, this requires some more logic to use the longname of the variable for display in the dashboard, but that is also doable.

@frazane @dnerini what do you think?

src/evalml/config.py
jonasbhend: will fix together with the forthcoming change in config below.

src/verification/__init__.py
frazane: Yes, I think "lt", "gt", etc. is a good proposition. I also support names without special characters.

dnerini: I like it too!

src/evalml/config.py
jonasbhend: Done

config/erroneous_config.yaml
dnerini, line 1: what is this about?

jonasbhend: oh good catch. This was to fix the pydantic schema to actually detect wrong configs. Not needed will remove.

dnerini (approved):
I tested the branch and it worked our very nicely! This seems ready to be merged to me!

PR #139, opened 2026-04-28
generalize the plot of meteograms to any location

cosunae:


workflow/Snakefile
```diff
@@ -138,21 +138,29 @@ rule experiment_all:
...
-            param=["T_2M", "SP_10M"],
-            region=["globe", "europe", "switzerland"],
-            showcase=EXPERIMENT_NAME,
+        (
+            expand(
+                rules.make_forecast_animation.output,
+                init_time=[t.strftime("%Y%m%d%H%M") for t in REFTIMES],
+                run_id=CANDIDATES,
+                param=["T_2M", "SP_10M"],
+                region=["globe", "europe", "switzerland"],
```
jonasbhend, line 147: Could we make these configurable as well? with corresponding entries in the showcase part of the config.

cosunae: yes, that is good idea

PR #140, opened 2026-04-28
Add option to extract the ensemble mean of a baseline

dnerini:


jonasbhend (approved):
Looking good, but I only had a quick look. Could you please add an example call for extracting the ensemble mean at the end of the script once you are sure it works?

workflow/scripts/data_extract_baseline.py
```diff
@@ -106,11 +109,54 @@ def extract(
...
+                if field.metadata("shortName") in params:
                     out.append(field)
+    return out.to_xarray(profile="grib")
 
-    out = out.to_xarray(profile="grib")
+
+def extract(
+    file: Path,
+    lead_times: list[int],
+    run_id: str,
```
clairemerker: I guess it's not too important since the function will not be used outside of this script, but it might be confusing to have a required argument that is not used if `ensemble_mean=True`

clairemerker (approved):
Looks good to me, I just had a very nitpicky comment you can ignore :) But I didn't test it and didn't verify output!

cosunae:
just one question, is it important to extract the mean and not the median? I am asking because the forecast archive files contain the median so the code would be 10x faster

PR #142, opened 2026-04-30
Configurable Stratification

jonasbhend:
**Please note: all of the changes implemented in this PR only affect the dashboard. The underlying verification files are untouched (i.e. still include the full stratification).**

## Changes
* add a new dashboard section to the main config
* allow users to omit stratifications from the dashboard to keep dashboard size manageable
* example config `interpolators-ich1.ymal` with seasonal stratification only reduces the size of the dashboard from 138 Mb to 9 Mb
* drop-downs are only shown for stratifications that are available, but all stratifications are indicated on the plot (see screen shot below)

image

### DIFF
--- a/config/forecasters-co1e.yaml
+++ b/config/forecasters-co1e.yaml
@@ -39,6 +39,12 @@ stratification:
     - alpensuedseite
   root: /scratch/mch/bhendj/regions/Prognoseregionen_LV95_20220517
 
+dashboard:
+  stratification:
+    # - init_hour
+    # - region
+    - season
+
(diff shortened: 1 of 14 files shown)


Louis-Frey:
Hi Jonas, nice new functionality and much needed! Thanks! 

I will test it once balfrin is again available and let you know.

dnerini (approved):
thank @jonasbhend this was very much needed. I tested the branch and was able to reduce the size of a dashboard from 900 MB to 20MB!

PR #144, opened 2026-05-01
Configurable showcases

cosunae:
The idea is to make the generation of showcases configurable. 
In the config we can now specify:
* whether we activate the generation of showcase animations and/or meteograms: 
* a list of locations
* the region, by name a custom configured extent/projection
* the parameters

An example
```
showcase:
  params:
    - T_2M
    - SP_10M
    - TOT_PREC
  meteograms: false
  animations: true
  regions:
    - europe
    - switzerland
    - name: alpine_arc
      extent: [-16.0, 25.0, 30.0, 65.0]
      projection: orthographic

  stations: [JUN, COV, GOR, WFJ, SAE, SAM, DAV, ZER, ANT, VSBAS, BRT, LTB, GOS, CEV, BIA]


```


workflow/Snakefile
```diff
@@ -138,21 +138,29 @@ rule experiment_all:
...
-            param=["T_2M", "SP_10M"],
-            region=["globe", "europe", "switzerland"],
-            showcase=EXPERIMENT_NAME,
+        (
+            expand(
+                rules.make_forecast_animation.output,
+                init_time=[t.strftime("%Y%m%d%H%M") for t in REFTIMES],
+                run_id=CANDIDATES,
+                param=["T_2M", "SP_10M"],
+                region=["globe", "europe", "switzerland"],
```
cosunae: Since I closed #139 in order to make a clean PR against main, I add here a link to follow a comment made:
https://github.com/MeteoSwiss/evalml/pull/139/changes/BASE..24848f1c23d328ac60027a1f31bf6d7dbae90b74#r3172572622
> Could we make these configurable as well? with corresponding entries in the showcase part of the config.

@jonasbhend could you please continue in this PR?

jonasbhend (changes requested):
This is looking good (from looking at it). Will test once balfrin is back online. Could you please include the example config in the code base (e.g. in interpolators-ich1.yaml), otherwise it might be lost.

config/interpolators-ich1.yaml
```diff
@@ -55,6 +55,22 @@ stratification:
...
   root: /scratch/mch/bhendj/regions/Prognoseregionen_LV95_20220517
 
+showcase:
+  params:
+    - T_2M
+    - SP_10M
+    - TOT_PREC
+  meteograms: false
+  animations: true
+  regions:
```
jonasbhend: Suggest we rename this to domains to avoid ambiguity with the 'regions' used in stratification.

cosunae: done

cosunae:
@jonasbhend did you have the chance to look at it?

config/interpolators-ich1.yaml
```diff
@@ -55,6 +55,22 @@ stratification:
...
+  meteograms: false
+  animations: true
+  domains:
+    - europe
+    - switzerland
+    - name: alpine_arc
+      extent: [3.0, 17.0, 43.5, 48.5]
+      projection: orthographic
+
+  stations: [JUN] #, COV, GOR, WFJ, SAE, SAM, DAV, ZER, ANT, VSBAS, BRT, LTB, GOS, CEV, BIA]
```
jonasbhend: I really like the idea of a showcase section in the config (and as @dnerini pointed out, we should probably do that for experiments too). What I am not sure is if a more hierarchical approach would be more user-friendly, e.g.:
```
showcase:
    meteograms:
        plot: true
        stations: [JUN, BAS, BER]
    animations:
        plot: true
        domains: ...
```
Then the question is, if you want params as top level (under showcase) or params per meteogram and animations separately. Easy enough to produce one after the other, so probably no need to replicate params in both.

fyi: @frazane

jonasbhend (changes requested):
Thanks @cosunae, really nice. I have a minor discussion point (not sure whether we want to do this at all) and there seems to be something wrong with the precipitation scaling in the example config (run as is for meteograms only):

Image

workflow/rules/plot.smk
```diff
@@ -100,13 +101,22 @@ rule plot_forecast_frame:
...
+            else ""
+        ),
         accu=lambda wc: int(RUN_CONFIGS[wc.run_id]["steps"].split("/")[2]),
     shell:
         """
         export ECCODES_DEFINITION_PATH=$(realpath .venv/share/eccodes-cosmo-resources/definitions)
         python {input.script} \
             --input {params.grib_out_dir}  --date {wildcards.init_time} --outfn {output[0]} \
             --param {wildcards.param} --leadtime {wildcards.leadtime} --region {wildcards.region} \
+            {params.region_extra} \
```
jonasbhend, line 119: AFAIK we are running a separate job for each region in the config. As showcases are already quite slow (due to the large volume of plots being generated), wouldn't it be more efficient to plot all the regions in one call, such that we have to load the data only once?

jonasbhend: Never mind, this is probably too hard due to the multiple outputs issue.

cosunae:
> Thanks @cosunae, really nice. I have a minor discussion point (not sure whether we want to do this at all) and there seems to be something wrong with the precipitation scaling in the example config (run as is for meteograms only):
> 
> Image

thanks @jonasbhend , not sure I understood what would you do here? 
It looks like CH2 shows very weird values probably due to the remapping of the grids and the nature of precip, but not sure what you would do with the scale?
Other examples look more reasonable... ?

202503270600_TOT_PREC_JUN

config/interpolators-ich1.yaml
cosunae: yes, that makes lot of sense. let me refactor the config alone the lines. Concerning the question, yes, do not have much opinion, probably more in favor of as you said not duplicate too much in order to not clutter much the config with replicated params

workflow/rules/plot.smk
cosunae: I think this makes lot of sense. I was already thinking alone those lines, for meteograms. 
Currently with snakemake we are generating rules per params x stations x runs. Each launch an individual script and re-loads the data. 
Similar to your comment, I think we could re-structure the granularity, and have python scripts that handle multiple regions, or multiple locations. And perhaps params.
But yes, I think that will be a considerable refactoring. I can take a look at after this PR?

cosunae:
@jonasbhend is there any other comment on this? 
I proposed a different PR for the optimization to unify the processing of regions and locations here: 
https://github.com/MeteoSwiss/evalml/pull/154

config/interpolators-ich1.yaml
```diff
@@ -60,16 +60,33 @@ showcase:
...
+    enabled: true
+    domains:
+      - europe
+      - switzerland
+      - name: alpine_arc
+        extent: [3.0, 17.0, 43.5, 48.5]
+        projection: orthographic
+thresholds:
+  TOT_PREC:
+    gt: [0.0, 0.001, 0.005]
```
jonasbhend, line 74: Please make sure to use the new units, TOT_PREC is now in mm (so thresholds should probably be `[0.0, 1.0, 5.0]`

jonasbhend (approved):
@cosunae Good to go after fixing the thresholds (due to change in units) for TOT_PREC in config

jonasbhend:
> > Thanks @cosunae, really nice. I have a minor discussion point (not sure whether we want to do this at all) and there seems to be something wrong with the precipitation scaling in the example config (run as is for meteograms only):
> > Image
> 
> thanks @jonasbhend , not sure I understood what would you do here? It looks like CH2 shows very weird values probably due to the remapping of the grids and the nature of precip, but not sure what you would do with the scale? Other examples look more reasonable... ?
> 
> 202503270600_TOT_PREC_JUN

No need to do anything about it in this PR.

jonasbhend:
Sorry @cosunae I inadvertently closed this (probably due to my heavily lagging connection yesterday). As said, this is good to go.

config/interpolators-ich1.yaml
cosunae: I see, where this change in the units came from? Because these values are in the main, should we do this in a different PR?

cosunae:
I had issues with a last run before merging, but in the end they were pointing to this https://github.com/MeteoSwiss/evalml/pull/158
so I will merge

PR #145, opened 2026-05-01
MRB-648 Scorecards for evalml

adestefani8:
## What this PR adds

This PR adds a new `report_scorecard` rule that renders a PNG comparing one run against one baseline.

The scorecard has:
- one row per `(variable × metric)`
- one column block per region
- one column per lead time inside each region block

Each cell encodes the row's metric as a relative difference between the two runs for that region and lead time: `(model − baseline) / |baseline| × 100`. 

Markers:
- blue → model better
- red → baseline better
- grey → `|diff|` below the neutral threshold (default 5%)
- grey `x` → missing or non-finite value

Scores:
- Supported scores: `RMSE`, `MAE`, `STDE`, `R2`, `ETS`, `POD`, `FAR`
- Score direction: `RMSE`, `MAE`, `STDE`, and `FAR` are lower-is-better; `R2`, `ETS`, and `POD` are higher-is-better.

Above the neutral threshold, dot area scales linearly with `|diff|%` and caps at `size_cap_pct` (default 30%).

scorecard_ICON-CH2-EPS_demo2

## Configuration

Configurable via `params` on the rule:

- `lead_times`: `"start/stop/step"` in hours.
- `regions`: regions to include as column blocks. If no region is specified, all regions are included.
- `variables`: `"VAR:M1,M2,..."` entries. Omit `:M1,M2,...` to use `all_metrics` for that variable.
  Metric names can also expand by prefix: requesting `ETS` includes all matching categorical scores, such as `ETS_gt_0p0`, `ETS_gt_0p001`, `ETS_gt_0p005`. 
  If no variable is specified, the script falls back to RMSE only for a default set of variables.

Other defaults (`season`, `init_hour`, metric settings, plot styling) live in the script's `cfg`.

## Plot layout

The plotting script makes a few automatic layout decisions:
- the longest region label is measured before rendering: `col_width` grows when necessary to prevent region header overlap, and the top margin/vertical separators adapt to the rendered header height
- the longest metric label is measured before rendering: variable labels keep a fixed gap from metric labels, and horizontal group separators start from the measured metric-label area
- the legend is centered on the scorecard area
- the no-data legend entry only appears when missing values are present

## TODOs

- [x] Expose main scorecard parameters in the evalml config
- [x] merge #160


dnerini:
looking good :)

image

workflow/rules/report.smk
```diff
@@ -46,3 +46,62 @@ rule report_experiment_dashboard:
...
+            "jura",
+        ],
+        variables=[
+            "U_10M:RMSE,MAE,STDE,CORR,R2",
+            "V_10M:RMSE,MAE,STDE,CORR,R2",
+            "T_2M:RMSE,MAE,STDE,CORR,R2",
+            "PMSL:RMSE,MAE,STDE,CORR,R2",
+            "TD_2M:RMSE,MAE,STDE,CORR,R2",
+            "TOT_PREC:RMSE,MAE,STDE,CORR,R2",
+        ],
```
jonasbhend: This is hard-coded in the rule. Would it make sense to expose this in the config (in particular given the expected rapid changes with new metrics. @cosunae in  https://github.com/MeteoSwiss/evalml/pull/144 has suggested a nice template for config 'sections', e.g.
```
runs:
...

experiment:
    dashboard:
        stratification: ...
    scorecard:
        regions:
            - all
            - mittelland
        scores: 
            U_10M: ["RMSE", "R2"]
            TOT_PREC: ["RMSE", "ETS_gt_0p0"]

showcases:
...
```

workflow/Snakefile
```diff
@@ -133,6 +133,12 @@ rule experiment_all:
             rules.verification_metrics_plot.output,
             experiment=EXPERIMENT_NAME,
         ),
+        expand(
+            rules.report_scorecard.output,
+            run_id=CANDIDATES,
```
jonasbhend: Currently we run this for all non-baseline runs against all baselines. It is configurable in the sense that you can copy-paste the config and remove runs / baselines that you don't want to be analysed. Is this enough?

workflow/rules/report.smk
```diff
@@ -46,3 +46,62 @@ rule report_experiment_dashboard:
...
+        script="workflow/scripts/report_scorecard.mo.py",
+        verif_run=lambda wc: EXPERIMENT_PARTICIPANTS[wc.run_id],
+        verif_baseline=lambda wc: EXPERIMENT_PARTICIPANTS[wc.baseline],
+    output:
+        report(
+            OUT_ROOT
+            / "results/{experiment}/scorecard_plots/{run_id}/scorecard_{baseline}.png",
+        ),
+    params:
+        lead_times="6/33/6",
```
jonasbhend: Also here, I suggest to make this configurable

jonasbhend (changes requested):
Very nice. I really like the scorecards. Great work. 

For future PRs, could you please add a short description of the changes (high-level overview) and - if necessary - also of the goals of the PR? That would be very helpful for the review.

As an additional suggestion, could we include the scorecard in the dashboard (I know we don't always want to produce it, but in case it is available it would be nice to include in a separate tab)?

workflow/rules/report.smk
jonasbhend: Also, I would suggest to only show ["RMSE", "R2"] by default. The other scores don't add much, or are even a bit dubious (such as correlation which cannot be generally converted to a skill score due to its support from -1 to 1). I know that Marco Arpagaus always wants to see STDE, but I think this is also based on a misunderstanding of what STDE is. STDE is only different from RMSE if we have 'systematic' domain-wide biases, which is usually not the case.

jonasbhend: With https://github.com/MeteoSwiss/evalml/pull/137 there are now additional scores available for categorical forecasts. ETS would be a candidate to include for TOT_PREC and wind in particular.

adestefani8: Hi Jonas, thanks for the suggestion. I included ETS in a recent commit, could you please have a look and let me know what you think?

jonasbhend: Hi Alberto. I really like the way the new scorecard looks. Super cool. 

I was wondering, if the referencing of the score shouldn't be made consistent with the  way these are represented in the scores file (`verif_aggregated.nc`). As such, I would the variables section above expect to look something like this:

```
       variables=[
            "T_2M:RMSE,MAE",
            "TOT_PREC:RMSE,MAE,ETS_gt_0p0",
        ],
``` 
That way, we can specifically include / exclude scores (and thresholds) on a per variable basiss.

workflow/scripts/report_scorecard.mo.py
```diff
@@ -0,0 +1,722 @@
...
+            "STDE",
+            "CORR",
+            "R2",
+            "ETS",
+            "POD",
+            "FAR",
+        ],  # every entry must appear in metric_directions
+        "metric_directions": {
+            "lower_is_better": ["RMSE", "MAE", "STDE", "FAR"],
+            "higher_is_better": ["CORR", "R2", "ETS", "POD"],
```
jonasbhend: As mentioned previously, I would not include CORR as an eligible metric to add to the scorecard. Especially so, as R2 is virtually the same.

workflow/scripts/report_scorecard.mo.py
```diff
@@ -0,0 +1,722 @@
...
+        vars_metrics = dict(_parse_var_metrics(s) for s in args.variable)
+    else:
+        vars_metrics = {
+            "U_10M": ["RMSE", "R2", "ETS"],
+            "V_10M": ["RMSE", "R2", "ETS"],
+            "T_2M": ["RMSE", "R2"],
+            "PMSL": ["RMSE", "R2"],
+            "TD_2M": ["RMSE", "R2"],
+            "TOT_PREC": ["RMSE", "R2", "ETS"],
+        }
```
jonasbhend: My comment from before about exposing the thresholds as well probably is better suited here. I.e. I would love to be able to specify what score AND what threshold to include on a per variable basis.

workflow/rules/report.smk
adestefani8: Hi thanks for the kind words!

This is actually already supported: that syntax in the `variables` param in `report.smk` gives
demoSyntax

That works thanks to this function in `report_scorecard`:
```
def _parse_var_metrics(spec: str):
    """Parse a 'VAR:M1,M2,...' (or 'VAR' alone) item into (var, [metrics])."""
    if ":" in spec:
        var, metrics = spec.split(":", 1)
        return var.strip(), [m.strip() for m in metrics.split(",") if m.strip()]
    return spec.strip(), None  # None → use all scores for this variable
```
It splits each entry first on `:` and then splits the metrics on `,`. So you can write just `ETS` to include all available thresholds, or list specific ones like `ETS_gt_0p0` as in your example.

workflow/scripts/report_scorecard.mo.py
adestefani8: Actually this block is just a hardcoded fallback that only triggers when `--variable` is not passed at all (e.g., running the script manually). If the snakemake rule always provides explicit `--variable` arguments, this code path is never hit in the workflow.

Honestly, I'm not sure this fallback is that useful now... A cleaner alternative might be to drop it and instead show everything present in the data when no `--variable` is given (and same thing with the regions).

jonasbhend: @adestefani8 Not sure I agree here. In the data we clearly also have metrics (e.g. BIAS) that cannot be converted to a skill score and thus for which it isn't easy to illustrate 'better than reference' as you do on the scorecard. Therefore, I suggest to keep a rather minimal set of metrics in case these are not specified, maybe just the RMSE to avoid overly complicated configuration.

workflow/Snakefile
dnerini: after e9c6752, multiple scorecards can be defined in the config file, example:
```
scorecard:
    nowcasting:
      baseline: INCA
      lead_times: "0/6/1"
      regions:
        - jura
        - mittelland
        - voralpen
        - alpennordhang
        - innerealpentaeler
        - alpensuedseite
      variables:
        - "U_10M:RMSE,R2,ETS"
        - "V_10M:RMSE,R2,ETS"
        - "T_2M:RMSE,R2,ETS"
        - "TOT_PREC:RMSE,R2,ETS"
    short_range:
      baseline: ICON-CH1-ctrl
      lead_times: "6/33/6"
      regions:
        - jura
        - mittelland
        - voralpen
        - alpennordhang
        - innerealpentaeler
        - alpensuedseite
      variables:
        - "U_10M:RMSE,R2,ETS"
        - "V_10M:RMSE,R2,ETS"
        - "T_2M:RMSE,R2,ETS"
        - "TOT_PREC:RMSE,R2,ETS"
    medium_range:
      baseline: ICON-CH2-ctrl
      lead_times: "24/120/24"
      regions:
        - jura
        - mittelland
        - voralpen
        - alpennordhang
        - innerealpentaeler
        - alpensuedseite
      variables:
        - "U_10M:RMSE,R2,ETS"
        - "V_10M:RMSE,R2,ETS"
        - "T_2M:RMSE,R2,ETS"
        - "TOT_PREC:RMSE,R2,ETS"
```

jonasbhend: Can we put this in an example config?

workflow/rules/report.smk
```diff
@@ -50,3 +50,47 @@ rule report_experiment_dashboard:
...
+rule report_scorecard:
+    input:
+        script="workflow/scripts/report_scorecard.mo.py",
+        verif_run=lambda wc: EXPERIMENT_PARTICIPANTS[wc.run_id],
+        verif_baseline=lambda wc: EXPERIMENT_PARTICIPANTS[
+            SCORECARD_CONFIGS[wc.scorecard_name]["baseline"]
+        ],
+    output:
+        report(
+            OUT_ROOT / "results/{experiment}/scorecards/{scorecard_name}/{run_id}.png",
```
jonasbhend: I suggest to expand the filename of the scorecard to facilitate identification when plots are copied/moved elsewhere, e.g. scorecard_{scorecard_name}_{experiment}_{run_id}.png ?

workflow/scripts/report_scorecard.mo.py
```diff
@@ -0,0 +1,713 @@
+import marimo
+
+__generated_with = "0.19.6"
+app = marimo.App(width="full")
```
jonasbhend: Do we need the marimo notebook functionality here. I see that @cosunae dropped using marimo in an earlier PR, so we may want to do the same here, no? Or is there a strong use-case for providing interactive funtionality (i.e. we envisage people fiddling with labels, etc.)

workflow/Snakefile
jonasbhend: already there, sorry tricked myself by looking at subset of commits

jonasbhend (approved):
i have just left two minor comments. Seems all good to me otherwise.

src/evalml/config.py
```diff
@@ -259,6 +259,44 @@ class AnimationsConfig(BaseModel):
...
+        description="Lead-time range as start/stop/step (hours).",
+    )
+    stratification: str = Field(
+        ...,
+        description="Dimension to use as scorecard columns (e.g. 'region').",
+    )
+    variables: List[str] = Field(
+        ...,
+        description="Variables and metrics as scorecard rows (VAR:M1,M2 format).",
+    )
```
adestefani8, line 285: Just as a side comment, right now the behavior is:

- `variables:` (null) → rejected by Pydantic
- key omitted → rejected by Pydantic
- `variables: []` → accepted, and falls through to the hard-coded RMSE-only set

Should we make the fallback more visible or leave it as it is?

workflow/scripts/report_scorecard.py
```diff
@@ -0,0 +1,622 @@
...
+        [
+            v
+            for v in diff_filtered.data_vars
+            if any(
+                s == m or s.startswith(f"{m}_")
+                for m in all_metrics
+                for s in [v.split(".")[-1]]
+            )
+        ]
+    ]
```
adestefani8: I think this `else` is unreachable: `cfg["variables"]` is always a non-empty dict since the `if args.variable / else` above sets it either from the parsed args or from the hard-coded RMSE-only fallback. So `cfg.get("variables")` should always be truthy.

workflow/scripts/report_scorecard.py
```diff
@@ -0,0 +1,622 @@
...
+_left_margin_col = figure.get("left_margin_in", 2.5) / col_width
+xlim_left = layout["metric_x"] - _left_margin_col
+
+_fixed = {dim: val for dim, val in _ALL.items() if dim != _strat}
+_init_raw = _fixed.get("init_hour")
+_title_parts = []
+if "season" in _fixed:
+    _title_parts.append(f"season={_fixed['season']}")
+if _init_raw is not None:
+    _title_parts.append(f"init={'all' if _init_raw == -999 else f'{_init_raw:02d}Z'}")
```
adestefani8: Here it would be nice to also show in the subtitle that we're aggregating over `region=all`.

src/evalml/config.py
dnerini: good point, I'll document the behavior more explicitly

workflow/scripts/report_scorecard.py
dnerini: good catch! I'll remove the dead code

dnerini:
let's merge this, thank you all for all the precious feedback and comments!

PR #146, opened 2026-05-04
Correct Units for TOT_PREC (MRB-820)

jonasbhend:
Precipitation is stored with units of `m` in anemoi datasets (e.g. ERA-5), and thus internally represented as `m` in anemoi models. The canonical units to display precipitation totals in meteorology are `kg m-2` or `mm`, thus a rescaling of precipitation outputs is necessary.

### Changes
* global postprocessors for all forecaster and interpolator outputs (accumulation of precipitation and rescaling to `mm`)
* pre-processors for grib inputs to interpolators to rescale back to `m`
* fixing precipitation units for analysis and baseline datasets

FYI: @hlasco


resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
```diff
@@ -5,6 +5,12 @@ input:
     - lam_0:
         grib:
           path: forecaster/20*
+          pre_processors:
+            - forward_transform_filter:
+                rescale:
+                  scale: 1000 # convert from kg m-2 to m
+                  offset: 0
+                  param: tp
```
jonasbhend, line 14: My understanding is, that the current version of the interpolator is not using TOT_PREC as an input. I would like to leave the pre-processor here, even if it is strictly not necessary. Just so we don't forget (assuming we evolve from here). What do you think, @MicheleCattaneo @frazane @dnerini

MicheleCattaneo: the interpolator does not use it as input, correct.
Leaving it there does not have an effect, but perhaps we can mention it in the comment?

resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
```diff
@@ -5,6 +5,12 @@ input:
     - lam_0:
         grib:
           path: forecaster/20*
+          pre_processors:
+            - forward_transform_filter:
+                rescale:
+                  scale: 1000 # convert from kg m-2 to m
```
MicheleCattaneo:
```suggestion
                  scale: 1000 # convert from kg m-2 to m (does not have an effect if interpolator does not use tp as prognostic variable)
```

resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
frazane: Yes I agree with Michele, fine to leave it there (if it still runs) but I'd add a comment.

resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
jonasbhend: fixed

resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
```diff
@@ -5,6 +5,13 @@ input:
     - lam_0:
         grib:
           path: forecaster/20*
+          pre_processors:
+            - forward_transform_filter:
+                # does not have an effect if the interpolator does not use tp as prognostic variable
+                rescale:
+                  scale: 1000 # convert from kg m-2 to m
```
Louis-Frey: Here you have `scale: 1000` in the pre_processor, in `sgm-interpolator-regional_trimedge.yaml` below you have `scale: 0.001` in the pre_processor (both in the interpolator). Shouldn't this be consistent?

src/data_input/__init__.py
```diff
@@ -150,7 +153,7 @@ def load_fct_data_from_grib(
         ## start precipitation here.
         diff = ds.TOT_PREC.diff("lead_time")
         min_diff = float(diff.min().compute())
-        if min_diff < -1e-4:  # -0.1 mm in precip-equivalent (units: m)
+        if min_diff < -0.1:  # TOT_PREC canonical units are mm
```
Louis-Frey, line 156: Thanks for thinking about this! ;-)

resources/inference/configs/sgm-interpolator-global_trimedge_multi.yaml
jonasbhend: good catch, thanks

Louis-Frey:
Thanks Jonas, I had a look, seems fine to me.

I tested all the ICON example configs, pinning anemoi-inference to `e369b1a90313e9701db13f63364a467aa281cf36` throughout (as in `interpolators-ich1.yaml`). They have all run without error. Not sure if this is how we want to keep testing though (pinning inference versions at random). 

I also tested a custom config on about two weeks of 6-hourly inits to see if the units in the lead-time plots behave as expected. This is the case:

RMSE_TOT_PREC

BIAS_TOT_PREC

Good to go from my side!

Louis-Frey (approved):

PR #147, opened 2026-05-05
Sticky column/row headers

cosunae:
**What this PR achieves:** 

* sticky column/row headers, so that scrolling keeps the headers in the frame
* zoom and interactive plot manipulation is individual
* more optimal rendenring by splitting into multiple SVG

 Instead of one monolithic Vega facet SVG (which makes sticky impossible without these hacks), render each (metric × param) cell as a separate small Vega chart embedded in an HTML <table>. 

The core idea: use a real HTML <table> instead of one big SVG. CSS position: sticky on `<th>` elements is native browser behavior — no JavaScript needed for headers to stay fixed.

Key points:

* border-collapse: separate  required for position: sticky to work on `<td>/<th>` elements
* z-index: 4 on the corner cell (it sits on top of both the column and row header layers)
* Each cell gets its own small Vega-Lite chart (280×160px), rendered independently
* Filter changes show/hide table rows and columns and re-embed only the visible cells
* Debounced updates (250ms) so rapid filter clicks don't spam re-renders
* _epoch counter aborts stale async renders if a newer one starts
* Legend is a zero-size point mark chart (the Vega-Lite trick for standalone legends)

Example (see how scrolling bars are in the middle of the page)
image


jonasbhend:
Very cool idea. I assume that zooming in to focus on a specific lead time range doesn't work across panels any more, right? Zooming is a feature that is probably worthwhile sacrificing...

frazane:
Could you share a sample dashboard generated with this code?

cosunae:
> Very cool idea. I assume that zooming in to focus on a specific lead time range doesn't work across panels any more, right? Zooming is a feature that is probably worthwhile sacrificing...

yes, exactly, The zooming here focus on a single panel (plot). I thought this was actually more convenient because it becomes more interactive (responsive), maybe I do not see exactly how you are normally using it. 

I will share an example, perhaps you can play with it.

jonasbhend:
> > Very cool idea. I assume that zooming in to focus on a specific lead time range doesn't work across panels any more, right? Zooming is a feature that is probably worthwhile sacrificing...
> 
> yes, exactly, The zooming here focus on a single panel (plot). I thought this was actually more convenient because it becomes more interactive (responsive), maybe I do not see exactly how you are normally using it.
> 
> I will share an example, perhaps you can play with it.

i was using the zooming to focus on a lead time range (usually the first day or so). With individual zooming, that is not anymore an option. Also, in a grid of plots, I would expect individual panels to share the axes (if these are aligned) and thus zooming to work across panels (on these axes). We don't do this for the y-axis, because the parameters are on different units, but x axis is the same, so I would expect all panels to have the same x-axis.

cosunae:
I think there was a solution to the zoom in, 
you can take a look at an example here: 
M:\zue-prod\fc_development\seamless\S-RUC\evaluation\MRB-585_dashboard_stage_E_interpolator.html

The zoom is more laggy than before since it needs to zoom all the panels, similar to what there is in main

jonasbhend:
I have another two minor suggestions:

* solid dots in the legend
* no blue lines to separate the header from the table of plots (just enough margin to visually set the plots apart from the headers)

cosunae:
What was done in the last commits: 
* solid dots in the legend
* no blue lines to separate the header from the table of plots (just enough margin to visually set the plots apart from the headers)
* increase the size of the main panel (table) of plots 
* buttom to hide filters (in order to gain more space for plot visualization once the selection is done)
* auto-scale the y axis

cosunae:
image

cosunae:
What was done in latest commits: 
* Fix memory issue (loading json multiple times in the table)
* auto-scale the y-axis while zoomin

cosunae:
@jonasbhend this would be ready from my side

config/forecasters-ich1.yaml
frazane, line 1: Has this change been tested also with categorical (threashold-based) scores?

frazane (approved):
Looks okay to me, just one quick question remaining

config/forecasters-ich1.yaml
cosunae: yes, this was a leftover from testing (I committed now to undo this commented out config for categorical scores). 
I was mainly testing on the interpolator-ich1, which included all categorical scores

frazane:
Good to be merged!

PR #150, opened 2026-05-07
Add lat lon information to ICON baselines

jonasbhend:
These changes have not been reconciled with main. Therefore another PR is necessary.


clairemerker (approved):
All good! Thanks for digging it out!

PR #151, opened 2026-05-11
Blacklist of forecast initializations

jonasbhend:
When running experiments with many initialization times, sometime individual initializations do not run successfully due to data problems. In order to deal with this problem, a blacklist of initialization times can be provided in the config to exclude from experiments.

## Summary of Changes
* adapted config schema and update generation of list of initialization times
* provide one example config with a blacklisted initialization


cosunae:
Looks good, if this is the case that some initializations have data problems, we could add that backlist to all configs, or rather have another config date_blacklist.yaml and make all the main configs point to it?

cosunae (approved):
looks good to me

PR #155, opened 2026-05-18
Side by side animations

cosunae:
Allows to generate side by side animations. 

The animations config (based on #144 ) allows to set the animation speed and the two panesl of the comparison (refering by label)

```
showcase:
  params:
    - T_2M
    - SP_10M
    - TOT_PREC
  meteograms:
    enabled: false
    stations: [JUN]
  animations:
    enabled: true
    domains:
      - europe
      - switzerland
      - globe
    speed: 10  # simulated hours per second
    runs:
      - Varda-Single
    comparisons:
      - left: Varda-Single
        right: KENDA-CH1
```

This PR implements the following functionality: 
* label all data (nested forecaster of the interpolator didnt have previously a label), experiments, baselines and truth
* add a plot_zarr_frame to support the plot generation of baselines and truth
 
This PR implements the following config syntax: 
* **include** statement in the yaml, to avoid config duplication (having to maintain showcases and dashboards in the same config)
*  **runs**  identify the labels that will be used for animation (single panel) animation
*  **comparisons**: identify the two labels that will be configured in the side by side comparison
* **speed**: animation speed as hourly simulated frames per second

TO BE MERGED AFTER #144


src/data_input/__init__.py
```diff
@@ -61,7 +61,8 @@ def load_analysis_data_from_zarr(
         "PMSL": "msl",
         "TOT_PREC": "tp",
     }
-    tot_prec_string = "TOT_PREC_6H" if min(np.diff(steps)) == 6 else "TOT_PREC_1H"
+    _diffs = np.diff(steps)
+    tot_prec_string = "TOT_PREC_6H" if len(_diffs) > 0 and min(_diffs) == 6 else "TOT_PREC_1H"
```
frazane: Could we add a comment explaining why we have to do this?

src/evalml/cli.py
frazane, line 1: Nice addition! I would add some tests for this since it can easily run in the GitHub CI.

src/plotting/compat.py
```diff
@@ -77,6 +77,60 @@ def load_state_from_grib(
     return state
 
 
+def load_state_from_zarr(
```
frazane, line 80: Would it make sense to wait for https://github.com/MeteoSwiss/evalml/pull/153 and read from GRIB files directly?

workflow/rules/plot.smk
```diff
@@ -153,8 +154,151 @@ rule make_forecast_animation:
...
+            --outfn {output} \
+            --param {wildcards.param} \
+            --leadtime {wildcards.leadtime} \
+            --region {wildcards.region} \
+            {params.region_extra} \
+            --accu {params.accu}
+        """
+
+
+rule make_zarr_animation:
```
frazane, line 248: Rule name should start with module name, see https://github.com/MeteoSwiss/evalml#workflow-development-guidelines

workflow/rules/plot.smk
```diff
@@ -153,8 +154,151 @@ rule make_forecast_animation:
...
+def get_zarr_leadtimes(wc):
+    """Get lead times for a zarr source, skipping step 0 for TOT_PREC."""
+    cfg = ZARR_SOURCES[wc.source_id]
+    step = cfg["step"]
+    total = cfg["total_hours"]
+    start = step  # always skip lead time 0 (no meaningful accumulation at t=0)
+    return [f"{i:03}" for i in range(start, total + 1, step)]
+
+
+rule plot_zarr_frame:
```
frazane, line 206: Logging is missing.

workflow/rules/plot.smk
```diff
@@ -153,8 +154,151 @@ rule make_forecast_animation:
...
+            / config["showcase"]["animations"].get("speed", 10.0)
+            * 100
+        ),
+    shell:
+        """
+        convert -delay {params.delay} -loop 0 {input} {output}
+        """
+
+
+rule make_comparison_animation:
```
frazane, line 278: Rule name does not follow guidelines, logging is missing.

frazane (commented):
Minor details. Looks very nice! Would you mind documenting somewhere your approach for wildcard constraints?

src/plotting/compat.py
cosunae: Yes, I thought so, but unfortunately the truth seems to stay in zarr, so I am assuming we need it still to be able to plot the truth?

https://github.com/MeteoSwiss/evalml/pull/153/changes#diff-c0f704baca2ade5ded6448745abbb4de19d01d0f2a7213b0c119fef7679c6cb7R44

src/evalml/cli.py
cosunae: I added some tests: 
https://github.com/MeteoSwiss/evalml/pull/155/changes/a291c59a61c021ff979810ddc5925c5aaf7d5002

I also added a load_yaml test (with the include), 
since this is writing a sample file, I am not sure this is what we should be doing in the unit tests, other if you have another idea to test it @frazane ?

https://github.com/MeteoSwiss/evalml/pull/155/changes#diff-217a439192bc495c66132a58a1bf7a92ee70f0eb9702c6b00f60644e726479b5R23

PR #157, opened 2026-05-20
Mrb 816 add inca hourly baseline to evalml

teobuz:
With this PR, we can add this to our config:

```
- baseline:
    label: INCA
    root: /store_new/mch/msclim/INCA
    steps: 0/6/1
```


src/data_input/__init__.py
```diff
@@ -7,13 +7,34 @@
...
+        _inca_lat_2d,
+        {"units": "degrees_north", "long_name": "latitude"},
+    ),
+    "lon": (
+        ("y", "x"),
+        _inca_lon_2d,
+        {"units": "degrees_east", "long_name": "longitude"},
+    ),
+}
+
```
jonasbhend: Maybe this could be simplified by using the coordinates (ch)x/y from the NetCDFs directly.

jonasbhend (commented):
Hi @teobuz and @dnerini

thanks for this. Seems to work from what I can tell, but I struggle with the interpolator config...

I was wondering if it wouldn't be easier to follow the grib loading template in that we:
1. compile a list of netcdf files to read from depending on frequency, parameter, initialization
2. use `xarray.open_mfdataset` to open all these at once
3. rename and add metadata if necessary

I know this is quite a refactoring from what we have now, so I would also be happy to postpone this to a later PR. Mainly I am wondering if there is something I have overlooked (missing parameters, etc.)

What do you think?

src/data_input/__init__.py
dnerini: the challenge is that _nan_array needs the grid shape upfront for fallback cases when file loading fails, but let's see if I can find a solution to that

jonasbhend: There is probably a deeper issue with NAs here. So far the pipeline fails if one of the input fields is not available (runs and baselines). I suggest we keep it that way. The whole point of the pipeline is that we have the exact same initializations from all the models and baselines for comparison, so we shouldn't ever have to pad with missing values. Then `xr.open_mfdataset` should work and one could even fail early if one of the expected inputs doesn't exist.

dnerini: ~I agree for missing reftimes, but we still need to pad with missing values in case of missing parameters (not all ICON variables are available in INCA).~

The issue with INCA is that it's relatively common to have missing variables or runs. If we were to apply the "pipeline fails if one of the inputs fields is not available" principle, I doubt we'll ever manage to complete an evaluation run

jonasbhend (approved):

src/data_input/__init__.py
jonasbhend: Yes, but I think we have to still do that, otherwise we lose all control over the results. In the end you could compare 5 initializations from INCA in winter to 365 from Varda throughout the year (I'm exagerating obviously). To facilitate, we could maintain a blacklist of initializations computed from INCA.

PR #158, opened 2026-05-21
Bugfix precip

cosunae:
Attempt to fix a bug with unit conversion of precip

### DIFF
--- a/src/data_input/__init__.py
+++ b/src/data_input/__init__.py
@@ -97,7 +97,11 @@ def load_analysis_data_from_zarr(
     )
 
     # change precipitation units from m to kg m-2
-    ds["TOT_PREC"] = ds["TOT_PREC"] * 1000  # convert precipitation units from m to mm
+    for prec_key in ("TOT_PREC_6H", "TOT_PREC_1H", "TOT_PREC"):
+        if prec_key in ds:
+            ds[prec_key] = (
+                ds[prec_key] * 1000
+            )  # convert precipitation units from m to mm
 
     # rename 'cell' dimension to 'values' (it's earthkit-data default for flattened spatial dim)
     if "cell" in ds.dims:
--- a/workflow/scripts/plot_forecast_frame.mo.py
+++ b/workflow/scripts/plot_forecast_frame.mo.py
@@ -144,33 +144,11 @@ def _(LOG, np):
             except Exception:
                 return arr - 273.15
 
-        def _ms_to_knots(arr):
-            # robust conversion with pint, fallback if dtype unsupported
-            try:
-                return (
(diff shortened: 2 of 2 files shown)


src/data_input/__init__.py
```diff
@@ -97,7 +97,11 @@ def load_analysis_data_from_zarr(
     )
 
     # change precipitation units from m to kg m-2
-    ds["TOT_PREC"] = ds["TOT_PREC"] * 1000  # convert precipitation units from m to mm
+    for prec_key in ("TOT_PREC_6H", "TOT_PREC_1H", "TOT_PREC"):
+        if prec_key in ds:
+            ds[prec_key] = (
+                ds[prec_key] * 1000
+            )  # convert precipitation units from m to mm
```
jonasbhend, line 104: AFAIK load_baseline_from_zarr at this stage only has a data_variable called TOT_PREC, as the 1h and 6h precipitation is renamed to the generic TOT_PREC (see definition of PARAMS_MAP). Or am I overlooking something?

cosunae: I think this was used by meteograms to load the truth (i.e. analysis) and I think the analysis has the TOT_PREC_1H and TOT_PREC_6H variables, but I might be wrong? I had a run crashing because TOT_PREC key didnt exists in ds

workflow/scripts/plot_forecast_frame.mo.py
```diff
@@ -194,7 +172,7 @@ def preprocess_field(param: str, state: dict):
             v = fields["V"]
             return np.sqrt(u**2 + v**2), "m/s"
         if param == "TOT_PREC":
-            return np.maximum(_m_to_mm(fields[param]), 0), "mm"
+            return np.maximum(fields[param], 0), "mm"
```
cosunae, line 175: I would have done something different, here we are making a strong assumption on the incoming units, which is fragile, since it will break when we change it in the inference. Instead, I would not assume what units field is coming in. I would call a function to_mm(field[param]), where the function checks the metadata() of the earthkit data object, and depending on the units we apply the conversion. 
But this requires a different loading of the data (in compat.py) which builds atm a custom dict state of numpy arrays, to just load an xarray (as we do with the baselines) and the move to earthkit-data. 
So I would leave the refactoring of this load of the grib files for plotting frames for later.

src/data_input/__init__.py
jonasbhend: Yep, the logic might be different in showcases. Obviously if you ask for TOT_PREC_6H (which we don't do in experiments) you would get uncorrected values. In that respect, your fix makes perfect sense.

jonasbhend (approved):

PR #160, opened 2026-05-22
Add ExperimentConfig for experiment workflow outputs

dnerini:
Apply same idea as in #144 , but for experiment outputs.

Note that this is a breaking change for configs!


jonasbhend (approved):
Thanks for implementing this. Much cleaner this way. This should also facilitate splitting up the config to reduce redundancy if we ever deem this to be necessary.

PR #163, opened 2026-05-26
Fix T_2M/TD_2M error by sourcing lead times 1–6 from previous INCA run

teobuz:
The INCA runs in the archive have a known bug where T_2M and TD_2M are incorrect at lead times 1–6 for runs initialized at full hours. This PR works around it by loading those steps from the run 10 minutes prior (at :50), while keeping step 0 from the current reftime.

The fix applies only to hourly frequency and the two affected parameters. A helper `_open_convert()` was extracted to reduce duplication in the loading logic.


src/data_input/__init__.py
```diff
@@ -416,6 +415,84 @@ def _nan_array(units: str) -> xr.DataArray:
...
+            )
+        except FileNotFoundError:
+            return fp, None
+        LOG.info("Reading %s", fp)
+        da = d[pfx]
+        u = da.attrs.get("units", "")
+        if u == "degrees C":
+            da = (da - ZERO_KELVIN).assign_attrs({**da.attrs, "units": "K"})
+        elif u == "mm/h":
+            da = da.assign_attrs({**da.attrs, "units": "kg m-2"})
```
frazane, line 441: Is this correct? We are converting from a rate to an accumulation. This particular conversion only works if values are for hourly accumulations. Is that always the case?

teobuz: Yes, good point, this work only for hourly accumulations, for 6h and 10min we have to do an additional conversion, I think it make sense to do it, so that we afterwards ready also for 10min accumulations. ANyway, we still have to compute the accumulations (now the algorithm just take the corresponding 10 min intensity). Probably make sense to open a new PR with accumulations and cooresponding units.

PR #164, opened 2026-05-27
Fix: read baselines from grib instead of zarr for meteograms

jonasbhend:
### Summary of changes
* pass on grib root instead of baseline zarr
* load_forecasts already knows how to handle grib root

### DIFF
--- a/workflow/scripts/plot_meteogram.py
+++ b/workflow/scripts/plot_meteogram.py
@@ -68,7 +68,7 @@ def main():
         action="append",
         type=str,
         default=[],
-        help="Path to baseline zarr data (repeatable).",
+        help="Path to baseline data root (repeatable).",
     )
     parser.add_argument(
         "--baseline_steps",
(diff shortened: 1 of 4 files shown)


jonasbhend:
I am now trying this out to make sure it works. If it does, could  you @cosunae  please add it to your PR (or we merge in sequence, once unify... is merged).

jonasbhend:
ok, this is now ready. Works with the interpolators-ich1.yaml config when enabling meteograms as well.

frazane:
Let's merge this! Thanks a lot @jonasbhend

workflow/scripts/plot_meteogram.py
```diff
@@ -36,7 +35,7 @@ def preprocess_ds(ds, param: str):
         }
         ds = ds.drop_vars(["U_10M", "V_10M"])
     if param == "SP":
-        ds[param] = ekm_wind.speed(ds.U, ds.V)
+        ds[param] = (ds.U**2 + ds.V**2) ** 0.5
```
cosunae, line 38: the use of ekm was implemented as part of a review, is it not working?

jonasbhend: no it is not working because earthkit wants a numpy array, but we are giving it a xarray DataArray. we could probably just change this to `ekm_wind.speed(ds.U.values, ds.V.values)` and then it should work. Should I revert that?

cosunae (approved):
lgtm

PR #165, opened 2026-05-29
fix uv requirements

cosunae:
**Issue with pytorch version:** 
uv 0.11.7 tightened PEP 440 local version matching — torch==2.8.0 no longer matches torch-2.8.0+cu126 on the PyTorch index, so uv fell back to PyPI and installed the cu128 build, which doesn't work with the CUDA 12.4 driver on Balfrin.

Fix: --index-strategy unsafe-best-match makes uv check all configured indices and pick the best matching version across all of them. With this flag, torch==2.8.0 resolves to 2.8.0+cu126 from the PyTorch index (which is a higher/better match than the PyPI 2.8.0 build), so the correct CUDA wheel is installed without needing to embed the +cu126 suffix explicitly in the requirements.

using uv 0.11.7 (x86_64-unknown-linux-gnu)


cosunae:
I would need some help testing this, with someone's else setup

jonasbhend:
Hi @cosunae, I have updated to uv 0.11.7 and have not had any problems. Is this still an issue? If not I suggest to close the PR.

PR #162, opened 2026-05-26
Migrate to earthkit v1.0 release candidate

frazane:
Migrates evalml to the first earthkit v1.0 release candidate. This meant moving data loading off meteodata-lab and onto earthkit-data, adapting to a GRIB encoding change in the new eccodes, and aligning coordinate and dimension names with what the new stack produces. Follow up PRs will do a more thorough refactoring of the data I/O and processing code - this is only to support earthkit 1.0.

**What changed**

- Updated dependencies in pyproject.toml and uv.lock: removed meteodata-lab and the old earthkit-plots, pinned the earthkit family (earthkit-data, earthkit-utils, earthkit-plots, earthkit-meteo, earthkit-geo) to release candidate versions, and bumped eccodes and eccodes-cosmo-resources-python. Also moved snakefmt to v2.0.
- Reworked GRIB reading in src/data_input on top of earthkit-data, replacing the meteodata-lab decoder. Includes fieldlist-to-xarray conversion driven by an xarray engine profile and de-accumulation handling for TOT_PREC.
- Standardized coordinate and dimension names across the codebase: latitude/longitude (previously lat/lon), step (previously lead_time), and valid_time (previously time). This touches data loading, spatial mapping in verification, plotting, the plot scripts, and the related tests.
- Adapted plotting to the new earthkit: the GRIB compatibility loader now goes through the shared data_input loader, and unit conversions and styles use earthkit-meteo and earthkit-plots.
- Fixed inference configs after a breaking change in the new eccodes: with the COSMO definitions active we can no longer encode both ICON and IFS GRIB files via shortName alone, so a modifiers/patches section now maps each variable to its paramId and shortName in the global inference configs.
- Removed legacy regional and trimedge inference configs that are no longer used.
- Added a README section on the migration, including a manual workaround to download and cache the eckit geo grid files for the ICON-CH grids, which earthkit cannot fetch automatically yet.

**Notes**

- earthkit v1.0rc is not final, so some rough edges remain. The ICON-CH grid cache step in the README is a manual workaround until the upstream download is fixed.
- The grib file globbing in plot_forecast_frame.py is a temporary fix (marked with a TODO) for anemoi-inference writing output filenames with unexpected formatting.
- **The plotting and data loading code urgently needs some care**. In this PR I focused on the minimal changes to make the code work with the new earthkit, but we need to do a larger refactor.

PR #167, opened 2026-06-01
Extract ICON baseline ensmean

dnerini:


jonasbhend (commented):
Cool. Looks straightforward to me. I will try now to see if it works as expected.

jonasbhend (approved):
Very nice. Very cool addition indeed. 

Using the median is a bit more tricky though, because we cannot aggregate on it. I.e. the 6hourly precipitation coming from the hourly median that is read through this PR will have a significant negative bias. So as soon as we make aggregation explicity, we should make sure that we read the corresponding median files (e.g. `..._median06h`) for aggregated parameters. 

Maybe best to add a warning / comment in the code so we don't forget.

dnerini:
Thanks for the review! Let's wait for https://github.com/MeteoSwiss/evalml/pull/162 before merging

PR #168, opened 2026-06-02
Retain sample size for further aggregation of strata

jonasbhend:
## Summary of changes
* retain a counter of samples for the different init_hour and season strata for further aggregation
* fix a bug in scaling of seasonal and per init_hour counts for the contingency table


dnerini (approved):
very nice, looking good to me!

PR #169, opened 2026-06-03
Introduce hash-based baseline IDs

dnerini:
Currently `baseline_id` is derived from the `label` field (or falls back to the stem of the `root` path). This means renaming the label silently changes the ID, breaking output path continuity. The `run_id` system already solves this problem: it uses a short SHA256 hash over the config fields that actually matter. Baseline IDs should follow the same contract.

This PR introduces hash-based baselines IDs, where `baseline_id` are computed as `baseline-{hash}` where `{hash}` is a 4-character SHA256 of the fields that define what the baseline *is* (`root`, `steps`, `member`), excluding the display-only `label` field.

(The PR also includes a few related minor quality-of-life improvements with respect to deprecated features and logging)


src/evalml/config.py
```diff
@@ -163,11 +163,9 @@ class InterpolatorConfig(RunConfig):
 class BaselineConfig(BaseModel):
     """Configuration for a single baseline to include in the verification."""
 
-    baseline_id: str | None = Field(
-        None,
-        min_length=1,
-        description="Deprecated compatibility field. Workflow baseline IDs are derived from the stem of `root`.",
-    )
+    # Fields excluded from ALL hashing (display metadata only).
+    HASH_EXCLUDE: ClassVar[FrozenSet[str]] = frozenset({"label"})
```
jonasbhend: i do not understand what this is doing here. It doesn't show up in the rendered config.schema.json and it is apparently overwritten in common.smk. Not saying this is wrong, just don't understand the rationale behind it. Thanks for clarification.

jonasbhend (approved):
Makes sense, thanks for implementing this. 

Ideally we would also do this for the truth, there we have the same problem. In addition, we recently identified with @Louis-Frey that none of our verification results (e.g. `verif.nc`) contain reference to the truth used, so will be overwritten when configs change (which is not ideal given that we may run the same config for analysis and stations). This should be fixed to. Happy to relegate to follow-up PR though.

dnerini:
Thanks Jonas, let's wait for #162 before merging this one

src/evalml/config.py
dnerini: good catch! there is no reason to have this field in config.py since the actual hash exclusion logic is implemented in common.smk, let me clean up this

PR #170, opened 2026-06-05
Power spectra plots

frazane:
Adds an optional power-spectra diagnostic to the experiment pipeline, so we can compare the effective resolution of model outputs against the truth and spot over-smoothing or spurious small-scale noise that point metrics like RMSE do not reveal.

The diagnostic computes 2D variance spectra (DCT by default, FFT optional) per participant and lead time, averages over initialisation times, and overlays each model against the truth with grid-resolution, effective-resolution and Nyquist reference lines.

**What changed**

- New src/spectra package: numpy-only spectral core (DCT/FFT with variance-conserving normalisation and radial averaging), ICON native-to-regular regridding with runtime grid detection, plus field extraction and IO helpers.
- Snakemake wiring (rules/spectra.smk): per-participant compute, aggregate over inits, overlay plot, gated behind an experiment target.
- Config: new experiment.spectra section (enable flag, method, lead times, variables, optional init-hour subset) with validation and regenerated JSON schema.
- Unit tests for the spectral core, regridding and IO.

**Notes**

- Disabled by default; opt in with experiment.spectra.enabled: true.
- ICON native grids are regridded to a regular grid before the transform; the eckit grid file is fetched automatically.

Example output, T_2M at +24h:

T_2M power spectrum at +24h


src/evalml/config.py
```diff
@@ -380,6 +380,57 @@ class ExperimentScorecardConfig(BaseModel):
...
+        default="dct",
+        description="Spectral method: 'dct' (default, recommended for LAM) or 'fft'.",
+    )
+    lead_times: List[int] = Field(
+        default_factory=list,
+        description="Representative lead times (hours) at which to compute spectra.",
+    )
+    variables: List[str] = Field(
+        default=["T_2M", "WIND_KE", "TOT_PREC"],
+        description="Spectra variables. Supported: T_2M, WIND_KE (from U/V_10M), TOT_PREC.",
```
jonasbhend, line 400: This now uses explicit aggregations, so this should probably be either TOT_PREC1 and/or TOT_PREC6

src/evalml/config.py
```diff
@@ -380,6 +380,57 @@ class ExperimentScorecardConfig(BaseModel):
...
+    )
+
+    @field_validator("variables")
+    @classmethod
+    def validate_variables(cls, v: List[str]) -> List[str]:
+        if not v:
+            raise ValueError(
+                "`variables` must list at least one variable when spectra is configured."
+            )
+        allowed = {"T_2M", "WIND_KE", "TOT_PREC"}
```
jonasbhend, line 414: same as above for TOT_PREC explicit aggregations

src/spectra/compute.py
```diff
@@ -0,0 +1,119 @@
...
+
+from spectra import core, io, regrid
+
+_PALETTE = ["#1f3b73", "#c1272d", "#1b7837", "#6a3d9a", "#e08214", "#4d4d4d"]
+
+
+def compute_source_spectra(ds: xr.Dataset, variables, lead_times, method, label):
+    """Compute spectra for one source (already-loaded native dataset).
+
+    Returns an x.Dataset with dims (variable, leadtime, wavenumber) and a shared
```
jonasbhend, line 20:
```suggestion
    Returns an xr.Dataset with dims (variable, leadtime, wavenumber) and a shared
```

src/spectra/compute.py
```diff
@@ -0,0 +1,119 @@
...
+    try:
+        stacked = xr.concat(datasets, dim="init")
+        agg = stacked.mean(dim="init", skipna=True)
+        agg.attrs = datasets[0].attrs
+        agg["wavelength"] = datasets[0]["wavelength"].copy()
+        agg.load()
+    finally:
+        for ds in datasets:
+            ds.close()
+    return agg
```
jonasbhend, line 76: In verification_aggregation.py we verify stratified by season (if configured). Could this be leveraged here, i.e. would it be easy to harmonize how we aggregate by initialization results to average results across the different sets of scores and metrics?

src/spectra/compute.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def plot_experiment_spectra(
+    truth_file, participant_files, out_dir, variables, lead_times
+):
```
jonasbhend, line 81: Not sure I agree on the way this module is organized. To me plotting definitely doesn't belong in compute, but would be better put in a separate sub-module. But as evalml is quite chaotic in this regard, I would be happy to defer this to a more general clean-up during the planned refactor.

jonasbhend: seeing that plot_power_spectra is defined in core, could this also go there?

src/spectra/core.py
```diff
@@ -0,0 +1,201 @@
...
+SPECTRUM_FUNCS = {"dct": dct_power_spectrum, "fft": fft_power_spectrum}
+
+
+def combined_spectrum(
+    spectrum_fn, fields_2d, dx_km, nbins=None, factor=1.0, oro_coeff=None
+):
+    """Sum the spectra of one or more component fields, scaled by `factor`.
+
+    For wind kinetic energy pass [u, v] and factor=0.5 -> 0.5*(|u_hat|^2+|v_hat|^2).
+    NaN bins are treated as zero in the sum so the common coordinate is preserved.
```
jonasbhend, line 110: Just out of curiosity, why would we want to do that instead of using spectra of wind_speed?

src/evalml/config.py
jonasbhend: Also, spectra for all the parameters that can be accessed with load_forecast/truth_data should be able to be computed, or does the computation require variable-specific configuration?

src/spectra/core.py
jonasbhend, line 1: I suggest to move all plotting to a separate sub-module for clarity.

src/spectra/io.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import xarray as xr
+
+# Logical spectra variable -> (component GRIB params, scale factor).
+# WIND_KE combines U/V as kinetic energy E = 0.5*(|u_hat|^2 + |v_hat|^2).
+VARIABLE_COMPONENTS: dict[str, tuple[list[str], float]] = {
+    "T_2M": (["T_2M"], 1.0),
+    "WIND_KE": (["U_10M", "V_10M"], 0.5),
+    "TOT_PREC": (["TOT_PREC"], 1.0),
+}
```
jonasbhend, line 15: This could be generalized to all available parameters (if computed directly) and only WIND_KE needs special treatment.

src/spectra/regrid.py
jonasbhend, line 1: Given that the regridding will be useful for other metrics, would it make sense to move into a separate module?

tests/unit/test_spectra_core.py
```diff
@@ -0,0 +1,103 @@
...
+        p = tmp_path / f"s_{init}.nc"
+        ds.to_netcdf(p)
+        return p
+
+    a = mk([np.nan, 2.0], 0)
+    b = mk([4.0, 4.0], 1)
+    agg = compute.aggregate_spectra([a, b])
+    np.testing.assert_allclose(
+        agg["power"].sel(variable="T_2M", leadtime=6).values, [4.0, 3.0]
+    )
```
jonasbhend, line 103: how could missing values, that we would want to exclude from aggregation, occur? Aren't all missing values real missing values that should result in a missing value in the aggregation? I am worried that we may get aggregated results based on very few non-missing samples without noticing.

workflow/rules/spectra.smk
```diff
@@ -0,0 +1,210 @@
...
+
+def spectra_participants():
+    """participant key -> aggregated spectra path (runs + baselines)."""
+    out = {}
+    for base in BASELINE_CONFIGS:
+        out[base] = OUT_ROOT / f"data/baselines/{base}/spectra_aggregated.nc"
+    for run_id, cfg in RUN_CONFIGS.items():
+        if cfg.get("_is_candidate", False):
+            out[run_id] = OUT_ROOT / f"data/runs/{run_id}/spectra_aggregated.nc"
+    return out
```
jonasbhend, line 40: are spectra participants different from experiment participants? Or should we consolidate and use only one?

jonasbhend (commented):
@frazane Thanks for pushing this. I have a few comments in particular around organization and generalization that should be easy to address. Good to go after that.

PR #141, opened 2026-04-29
Add Sphinx docs with GitHub Pages deploy

clairemerker:
Claude-generated docs for evalml.

PR #175, opened 2026-06-07
Fix showcases

dnerini:
This PR fixes a few recent issues related to the showcase command.

- #154 changed `plot_forecast_frame` to output frames for all regions in a single job (to reduce data I/O). However, `make_forecast_animation` was not updated accordingly: its input lambda expanded `rules.plot_forecast_frame.output`, which already has {region} resolved for all regions, causing the `region=wc.region` argument to be silently ignored. As a result, every animation job collected frames from all regions and passed them all to convert, producing incorrect GIFs.
  - fix: 9fdefc8

- #162 with eartkit-plots 1.0 something broke in drawing the colormap for precip. For the fix I relied heavily on Claude and will need to be double checked. In essence, for `TOT_PREC`, it seems that because the hex color list was passed as a raw kwarg, it was setting `no_style=True`, after which `configure_style` had already popped colors → matplotlib used viridis. Fixed by embedding colors in the `Style` object rather than passing them as kwargs; `get_style` now detects a pre-built `Colormap` and uses earthkit's vmin/vmax path (which handles `Colormap` objects), injecting `levels` directly into `style._kwargs` to bypass interception. 

   - also, `extend` moved into each preset in `colormap_defaults.py` (precipitation uses "max" since values are clipped to ≥ 0)
   - fix: da014ab

- consolidate region domains: 671248e

- SLURM jobs timing out because `plot_forecast_frame` tasks were being grouped into batches of 8 per SLURM job, but `--cores 4` limited the inner snakemake to run them sequentially. With 8 tasks running sequentially at ~9 minutes each, they'd exceed the 20-minute SLURM time limit. By removing the batching, each task now runs independently in its own job, allowing them to execute in parallel.
  - fix: f9ea85c

### DIFF
--- a/config/interpolators-ich1.yaml
+++ b/config/interpolators-ich1.yaml
@@ -116,11 +116,10 @@ showcase:
   animations:
     enabled: true
     domains:
+      - globe
       - europe
+      - icon-ch1
       - switzerland
-      - name: alpine_arc
-        extent: [3.0, 17.0, 43.5, 48.5]
-        projection: orthographic
 
 locations:
   output_root: output/
--- a/config/interpolators-ich1.yaml
+++ b/config/interpolators-ich1.yaml
@@ -118,7 +118,8 @@ showcase:
     domains:
       - globe
       - europe
-      - icon-ch1
+      - alps
+      - icon-ch
       - switzerland
 
 locations:
--- a/config/interpolators-ich1.yaml
+++ b/config/interpolators-ich1.yaml
@@ -116,9 +116,9 @@ showcase:
   animations:
     enabled: true
     domains:
-      - globe
+      # - globe
       - europe
-      - alps
+      # - alps
       - icon-ch
       - switzerland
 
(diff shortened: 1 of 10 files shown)


config/interpolators-ich1.yaml
```diff
@@ -116,11 +116,11 @@ showcase:
...
     enabled: true
     domains:
+      # - globe
       - europe
+      # - alps
+      - icon-ch
       - switzerland
-      - name: alpine_arc
-        extent: [3.0, 17.0, 43.5, 48.5]
-        projection: orthographic
```
jonasbhend, line 123: We seem to be doing this a lot. Introduce a new feature but then we don't keep it in an example config. Could we do better here? I hope it is basically in the schema definition, but I struggle withe readability of the schema and plain trial-and-error seems very inefficient. Instead I think we should have a help page on all of the config options.

jonasbhend (approved):
Thanks Daniele for spotting this. This looks good from a visual inspection. I can run the showcase to confirm, but would rather not as it is still very slow (we probably need to fix that).

config/interpolators-ich1.yaml
dnerini: I agree! wouldn't #141 provide that?

jonasbhend: would be good to check. I'm on it.

Issue #129, opened 2026-04-01
Adding new variables breaks inference


frazane:
Whenever we add new inputs or outputs to our models, there is a high chance that we break inference because some components of our inference pipeline (namely the metadata patches and the GRIB templates) are hardcoded with a limited set of variables. This forces us to merge hotfixes such as #128, which is not ideal. 

We should:

* change the GRIB templating approach so that we don't need to specify the list of params
* create variables metadata patches starting from the anemoi datasets used for training (the current patches were derived from some checkpoints), which contain all the possible variables we could need in inference

PR #178, opened 2026-06-09
Add diagnostic windgust (support for multi-datasets)

radiradev:
Support running the multi-decoder with a VMAX_10M on REAL-CH1:

had to pin anemoi-inference (radiradev/anemoi-inference@fix/empty-input-propagate-date) with two fixes. (still unsure if the padding one is needed). 

Also unsure what is the best way here to deal with the templates and their patches #129 

image


resources/inference/configs/sgm-multidataset-forecaster-windgust-ich1.yaml
```diff
@@ -0,0 +1,58 @@
...
+            samples: resources/templates_index_ifs.yaml
+          post_processors:
+            - extract_mask: # removes LAM points
+                mask: "lam_0/cutout_mask"
+                as_slice: true
+                inverse: true
+            - assign_mask: # fill local/global overlapping points with nan
+                mask: "global/cutout_mask"
+  # Dianostic decoder on ICON grid only
+  realv2:
```
radiradev, line 43: `realv2` should be called something else like `secondary_decoder`

resources/inference/configs/sgm-multidataset-forecaster-windgust-ich1.yaml
```diff
@@ -0,0 +1,58 @@
...
+    grib:
+      path: grib/realv2-{date}{time:04}_{step:03}.grib
```
radiradev, line 45: here too

PR #179, opened 2026-06-09
Source of truth in verification results

jonasbhend:
Currently, we don't record the truth used to produce verification results. Therefore, if you recompute evaluation with a different truth (e.g. analysis / stations), the intermediate verification files (currently `verif.nc` and `verif_aggregated.nc`) will be overwritten. To avoid this, we propose to make intermediate verification results truth-specific. Also, this may open up the possibility to run evaluation experiments with multiple truth data sources in the future.

### Summary of changes
* introduce truth (label) in rules producing intermediate verification files
* record truth label and root in NetCDF metadata (convenience)

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 
 CONFIG_ROOT = Path("config").resolve()
 OUT_ROOT = Path(config["locations"]["output_root"])
+TRUTH_LABEL = config["truth"]["label"]
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M"
 HASH_LENGTH = 4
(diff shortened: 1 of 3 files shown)


clairemerker (approved):
Good extension, thanks! :) The truth label used will be the one already set in the configs, right?

dnerini (commented):
Thanks! I like this addition, but the approach we used elsewhere is based on hashing the relevant parameters instead of using labels, see also  https://github.com/MeteoSwiss/evalml/pull/169 

Did you consider doing the same here?

jonasbhend:
> Thanks! I like this addition, but the approach we used elsewhere is based on hashing the relevant parameters instead of using labels, see also #169
> 
> Did you consider doing the same here?

Yes of course. I know that the label is not stable (pun intended), but I was trying to accommodate the use case that we actually work with the data more interactively (e.g. copy a netcdf file somewhere to plot scores for a presentation, or whatever). While writing this response, I see now that I didn't think this through, as we need evalml to figure out what the directory names mean anyway (run_id's are hashed), so might as well hash the truth too. Will get to it...

dnerini (approved):
looking good!

PR #182, opened 2026-06-12
Support MLflow model registry checkpoints

lclanzi:
Support MLflow model registry URLs as checkpoint references, in addition to run-based URLs.

## Previously supported (run-based):
checkpoint: https://service.meteoswiss.ch/mlstore#/experiments/42/runs/abc123...

## Now also supported (model registry):


dnerini (approved):
very good, nice and clean implementation!

dnerini:
sorry @lclanzi , I realized I was too quick in the review above... I reverted the merge and we'll need to put some more work into this. The main issue is that the code needs to download the checkpoint from a registered model, as opposed to only look up the the checkpoint path from a logged run.

PR #183, opened 2026-06-12
Retrieve station observations from MeteoSwiss DWH

jonasbhend:
AI-assisted implementation!

Adds a new verification truth source that retrieves SwissMetNet (SMN) surface observations live from the MeteoSwiss data warehouse (DWH) via the jretrievedwh.py client. This lets us verify model forecasts against station observations pulled on demand.

The implementation mirrors the approach already used in [anemoi-plugins-meteoswiss add-synop-dwh-source](https://github.com/MeteoSwiss/anemoi-plugins-meteoswiss/tree/add-synop-dwh-source/src/anemoi_plugins_meteoswiss/transform/sources/synop_dwh), and produces an xr.Dataset in the exact same shape as load_obs_data_from_peakweather, so the downstream spatial mapping (map_forecast_to_truth) and verify() work unchanged.

### Improvements:
- [x] check all valid times are present in `_select_valid_times`
- [x] fail fast when access is not available
- [x] only retrieve necessary data (increment is hardcoded to 60 min)
- [x] use varda's own jretrieve credentials

### Summary of changes
* wire-up data retrieval with jretrievedwh.py client
* provide authorization with specific credentials via .jretrievedwh-conf.prod.py
* add jretrieve markers in config for different retrieval options, including:
    * by measurement category group ids (default)
    * by station shortname
    * by bounding box (lon/lat)
* use DWH metadata to infer station locations for meteogram plotting (e.g. with analysis as truth)
* remove peakweather from dependencies and as a data source

### Examples
202503010000_T_2M_BER

202503010000_T_2M_BER

file:///M:/zue-prod/fc_development/seamless/S-RUC/evaluation/MRB-899_dashboard.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,33 @@ You can then run it with:
 evalml experiment path/to/experiment/config.yaml --report
 ```
 
+### Truth sources
+
+The `truth.root` value selects how the ground truth is loaded:
+
+- **Analysis Zarr** — a path ending in `.zarr` (anemoi analysis dataset).
+- **PeakWeather** — a path containing `peakweather` (SwissMetNet station obs from Hugging Face).
+- **DWH / jretrievedwh** — a `jretrievedwh:` marker string fetching SwissMetNet (SNM)
+  surface observations live from the MeteoSwiss data warehouse. Variables are mapped to
+  ICON names in SI units (temperatures in K, pressure in Pa, precipitation as the hourly
+  sum); wind `U_10M`/`V_10M` are derived from speed + direction.
+
+  Marker syntax (station selection is required; pick one of group/locations/bbox):
+
+  ```yaml
+  truth:
+    label: SwissMetNet (DWH)
+    root: jretrievedwh:SwissMetNet                  # stn_group (default group)
+    # root: jretrievedwh:locations=ARO,KLO,LUG      # explicit nat_abbr list
+    # root: jretrievedwh:bbox=45.8,47.8,5.9,10.5    # minlat,maxlat,minlon,maxlon
+    # append ;stage=devt to target a non-prod DWH stage (prod|depl|devt)
+  ```
+
+  **Prerequisites:** `jretrievedwh.py` must be on `$PATH` (e.g.
+  `/oprusers/osm/opr.inn/bin`) and `$OPR_HOME` set with a readable
+  `.jretrievedwh-conf.<stage>.py` conf file. No data is pre-downloaded — the obs are
+  queried at verification time.
+
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -166,10 +166,10 @@ The `truth.root` value selects how the ground truth is loaded:
     # append ;stage=devt to target a non-prod DWH stage (prod|depl|devt)
   ```
 
-  **Prerequisites:** `jretrievedwh.py` must be on `$PATH` (e.g.
-  `/oprusers/osm/opr.inn/bin`) and `$OPR_HOME` set with a readable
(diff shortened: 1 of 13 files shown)


src/data_input/__init__.py
```diff
@@ -264,71 +270,134 @@ def load_forecast_data_from_grib(files: list[Path], params: list[str]) -> xr.Dat
     return ds
 
 
-def load_obs_data_from_peakweather(
-    root, reftime: datetime, steps: list[int], params: list[str], freq: str = "1h"
-) -> xr.Dataset:
-    """Load PeakWeather station observations into an xarray Dataset.
+DWH_PARAM_MAP = {
```
dnerini: the global `DWH_` variables are used only by `load_obs_data_from_jretrieve`, I suggest moving them at the top of that function

README.md
```diff
@@ -145,6 +145,31 @@ You can then run it with:
...
+
+  ```yaml
+  truth:
+    label: SwissMetNet
+    root: jretrievedwh:1,2                          # stn_group_id (default)
+    # root: jretrievedwh:locations=ARO,KLO,LUG      # explicit nat_abbr list
+    # root: jretrievedwh:bbox=45.8,47.8,5.9,10.5    # minlat,maxlat,minlon,maxlon
+  ```
+
+  **Prerequisites:** `jretrievedwh.py` must be on `$PATH` (falls back to
```
dnerini, line 168: would it be possible to package this as a python package that we add as project dependency?

README.md
```diff
@@ -175,6 +200,25 @@ valid for 30 days. Every training or evaluation run within this period automatic
...
+
+To use `jretrievedwh:` as a truth source, provide a client ID and secret for
+the MeteoSwiss service account. Set them as environment variables:
+
+```bash
+export JRETRIEVE_CLIENT_ID=your-client-id
+export JRETRIEVE_CLIENT_SECRET=your-client-secret
+```
+
+or place them in a `.env` file at the project root (next to `.jretrievedwh-conf.prod.py`):
```
dnerini, line 213: would it help to include a `.env.template` ?

dnerini (commented):
Fantastic contribution, many thanks for this! I'm having only a first look, but this PR looks already quite mature! I have a few initial very minor comments. I'm also testing in myself and so far everything is running very smoothly.

I have perhaps one question, how hard would be to add another importer for obs later on? maybe something like `load_obs_data_from_ogd`? would the current code structure allow for that?

jonasbhend:
> Fantastic contribution, many thanks for this! I'm having only a first look, but this PR looks already quite mature! I have a few initial very minor comments. I'm also testing in myself and so far everything is running very smoothly.
> 
> I have perhaps one question, how hard would be to add another importer for obs later on? maybe something like `load_obs_data_from_ogd`? would the current code structure allow for that?

I would say this should work quite seamlessly, as long as the output can be wrangled in the same form as what we get from jretrieve. One of the issues I see with OGD, is that there you currently access station data by station and not by time, so probably is not very useful to our use-case with repeated requests per initialization time.

README.md
jonasbhend: @clairemerker I think I lack a bit of insight on what this does and how much effort this would be. I agree though that the current solution is 'hacky'

clairemerker: Hm... It's part of this repo: https://service.meteoswiss.ch/git/databrokerandcustomerinterfaces/jretrieve 
So if anything we would need to get if from there? I wouldn't create our own package for it...

dnerini: OK I see, thank you. I suggest to leave this as it for now. We can look into making this more robust in the future.

dnerini (approved):

PR #186, opened 2026-06-17
Feat/support checkpoint from model registry

lclanzi:
Support MLflow model registry URLs as checkpoint references, in addition to run-based URLs.

### Previously supported (run-based):
checkpoint: https://service.meteoswiss.ch/mlstore#/experiments/42/runs/abc123...

### Now also supported (model registry):

### Additional notes
If the model version is not provided, e.g. 
checkpoint: https://service.meteoswiss.ch/mlstore#/models/sruc-m-1-forecaster
the latest model version is used.

Moreover, when the checkpoint references a model in the model registry, the artifacts (i.e. inference-last.ckpt) is downloaded locally instead of being read from a local path.


dnerini (approved):
nice work, thank you! I tested it (see new config) and worked out nicely :)

PR #187, opened 2026-06-17
Fix inference env target for compilation

MicheleCattaneo:
The inference environment target for compilation was pointing to the global evalml venv. This PR fixes the target to be that used for inference.


dnerini (approved):
Nice catch! I wonder if this will have any positive impact... did you notice anything on your side?

PR #189, opened 2026-06-17
Make environment squashing optional

lclanzi:
Add a `skip_env_squashfs` boolean flag to RunConfig that allows bypassing squashfs image creation and using the virtual environment directly. I believe this can be useful during development or quick iterations where the ~5–10 min mksquashfs step is a bottleneck.

Config example:
```runs:
  - forecaster:
      checkpoint: https://servicedepl.meteoswiss.ch/mlstore#/experiments/409/runs/b30acf68520a4bbd8324c44666561696
      label: stage_C_icon_1km
      steps: 0/120/6
      skip_env_squashfs: true
      config: resources/inference/configs/sgm-forecaster-global-ich1-oper.yaml
```

### DIFF
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -101,10 +101,10 @@ class RunConfig(BaseModel):
         False,
         description="If true, the ECCODES_DEFINITION_PATH environment variable will not be set to the COSMO local definitions.",
     )
-    skip_env_squashfs: bool = Field(
-        False,
-        description="If true, skip creating a squashfs image and activate the virtual environment directly. "
-        "Useful when squashfs creation is too slow and a quick iteration is preferred.",
+    squash_venv: bool = Field(
+        True,
+        description="If true (default), package the virtual environment as a squashfs image before running inference. "
+        "Set to false to activate the virtual environment directly and skip the squashfs build step.",
     )
 
     config: Dict[str, Any] | str
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -101,6 +101,11 @@ class RunConfig(BaseModel):
         False,
         description="If true, the ECCODES_DEFINITION_PATH environment variable will not be set to the COSMO local definitions.",
     )
+    skip_env_squashfs: bool = Field(
+        False,
+        description="If true, skip creating a squashfs image and activate the virtual environment directly. "
+        "Useful when squashfs creation is too slow and a quick iteration is preferred.",
+    )
 
     config: Dict[str, Any] | str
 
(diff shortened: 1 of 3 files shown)


workflow/rules/inference.smk
```diff
@@ -321,7 +325,32 @@ rule inference_execute:
...
+                        --unbuffered \
+                        --partition={resources.slurm_partition} \
+                        --cpus-per-task={resources.cpus_per_task} \
+                        --mem-per-cpu={resources.mem_mb_per_cpu} \
+                        --time={resources.runtime} \
+                        --gres={resources.gres} \
+                        --ntasks={resources.ntasks} \
+                        anemoi-inference run config.yaml "${{CMD_ARGS[@]}}"
+                    '
+            fi
```
dnerini, line 336: mmmh I'm not really happy with all this code duplication. Can we try avoid it?

src/evalml/config.py
```diff
@@ -101,6 +101,11 @@ class RunConfig(BaseModel):
         False,
         description="If true, the ECCODES_DEFINITION_PATH environment variable will not be set to the COSMO local definitions.",
     )
+    skip_env_squashfs: bool = Field(
```
dnerini: I wonder if we should phrase it in a "positive" sense and more in the uenv sense: using e.g. a `--build-uenv` flag that is true by default?

dnerini (commented):
thanks for the nice contribution! I can see how this is really useful for debugging, although I'm slightly concerned that the option will be used a lot once we realize that it makes the code faster (the squash-fs step follows best practices from CSCS concerning python virtual envs)

workflow/rules/inference.smk
lclanzi: Updated in latest commit

lclanzi:
> thanks for the nice contribution! I can see how this is really useful for debugging, although I'm slightly concerned that the option will be used a lot once we realize that it makes the code faster (the squash-fs step follows best practices from CSCS concerning python virtual envs)

I see, that's a valid point. If this is the case, feel free to reject the PR

dnerini:
> > thanks for the nice contribution! I can see how this is really useful for debugging, although I'm slightly concerned that the option will be used a lot once we realize that it makes the code faster (the squash-fs step follows best practices from CSCS concerning python virtual envs)
> 
> I see, that's a valid point. If this is the case, feel free to reject the PR

maybe it's enough to hide it well ;) jokes asides, I really think this can be useful, so good documentation should be enough

src/evalml/config.py
lclanzi: Updated in latest commit

jonasbhend (approved):
Looking good to me. Tested it. Should be useful for speeding up integration tests.

PR #190, opened 2026-06-17
Add support for aifs-single-1.1

dnerini:


resources/inference/templates/templates_index_global-ea.yaml
frazane, line 1: I am not sure we need to even define GRIB templates. Have you tried without?

config/aifs-single.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - U_10M
+    - V_10M
+    - TOT_PREC
+  dashboard:
+    stratification:
+      # - region
+      # - init_hour
+      - season
+  stratification:
+    regions: []
```
frazane, line 45: What is the current behavior w.r.t. regions stratification? For instance here we pass an empty list - it means we compute the verification globally or only over the Alpine domain?

resources/inference/configs/aifs-single-forecaster.yaml
```diff
@@ -0,0 +1,22 @@
...
+      accumulations:
+        - tp
+
+output:
+  grib:
+    path: grib/{date}{time:04}_{step:03}.grib
+    templates:
+      samples: resources/templates_index_global-ea.yaml
+
+write_initial_state: false
```
frazane: Why is that? Could still be nice to see especially when comparing against observations.

resources/inference/templates/templates_index_global-ea.yaml
dnerini: I was also confused, but it was failing without...

config/aifs-single.yaml
dnerini: the "all" region is currently hardcoded, see https://github.com/MeteoSwiss/evalml/blob/cc5836625fee7a6f97195788190ac362d24ed3df/src/verification/__init__.py#L49

resources/inference/configs/aifs-single-forecaster.yaml
dnerini: mmmh it might be a leftover from when I was debugging the grib template issue, I'll try to remove it!

src/data_input/__init__.py
```diff
@@ -69,7 +83,12 @@ def load_analysis_data_from_zarr(
     PARAMS_MAP_COSMO1 = {
         v: v.replace("TOT_PREC", tot_prec_string) for v in PARAMS_MAP_COSMO2.keys()
     }
-    PARAMS_MAP = PARAMS_MAP_COSMO2 if "co2" in root.name else PARAMS_MAP_COSMO1
+    USE_IFS_NAMES = {"-co2-", "-ea-"}
```
MicheleCattaneo:
```suggestion
    USE_IFS_NAMES = {"-co2-", "-ea-", "ifsnames"}
```
In case, for any reason, we use rea-l with the ifnames?

MicheleCattaneo (commented):
It seem to be looking good. I could only spot one thing so far, made a minor comment

resources/inference/templates/templates_index_global-ea.yaml
dnerini: indeed it works without! thanks!

config/aifs-single.yaml
jonasbhend, line 1: i suggest to include AIFS in the varda-single config instead.

src/data_input/__init__.py
```diff
@@ -10,6 +10,20 @@
...
+_IFS_TO_COSMO = {
+    "tp": "TOT_PREC",
+    "msl": "PMSL",
+    "10u": "U_10M",
+    "10v": "V_10M",
+    "2t": "T_2M",
+    "2d": "TD_2M",
+    "sp": "PS",
+    "lsm": "FR_LAND",
+}
```
jonasbhend: From what I can tell, this is the same as `PARAMS_MAP_COSMO2` below. I suggest to simplify by finding the right conditions when to apply `_IFS_TO_COSMO` / `_COSMO_TO_IFS` and when to use the parameters unchanged (because they already follow COSMO naming). Having said that, I much prefer your naming and approach with specifying the map on top. `PARAMS_MAP_COSMO1/2` really doesn't make sense anymore.

src/evalml/config.py
```diff
@@ -377,12 +377,12 @@ class Stratification(BaseModel):
...
-        description="List of region names for stratification.",
+        default_factory=list,
+        description="List of region names for stratification. Empty list means no spatial stratification.",
     )
-    root: str = Field(
-        ...,
-        description="Root directory where the region shapefiles are stored.",
+    root: Optional[str] = Field(
+        None,
+        description="Root directory where the region shapefiles are stored. Required when regions is non-empty.",
```
jonasbhend, line 385: Thanks for cleaning this up.

jonasbhend (approved):
Works for me, but only tested the experiment. 

At some point we have to address the issue that we cannot use arbitrary runs for 'baselines' in the scorecards.

config/aifs-single.yaml
dnerini: why? I thought it could be interesting to maintain a selection of models in separate files, which would then make it easier for anyone to combine them as necessary into new configs...

src/data_input/__init__.py
frazane: Generally speaking I would get rid of any reference to "COSMO-2" dataset since we no longer use it.

dnerini: 945f1e9

PR #191, opened 2026-06-18
Add lapse-rate correction for temperature

jonasbhend:
lapse-rate correction works for 
- [x] Varda-Single
- [x] ICON-CH1-EPS
- [x] ICON-CH2-EPS
- [x] INCA
- [x] REA-L-CH1
- [x] KENDA-CH1
- [x] SwissMetNet
- [x] experiments
- [x] meteogram from showcases
- [x] Solve issue with ICON / KENDA / REA-L-CH1 grids

### Summary of changes
* retrieve altitude information from FIS (surface geopotential) for KENDA-CH1
* retrieve altitude information for ICON-CH1/2-EPS and Varda from ICON constants files on balfrin
* retrieve altitude information for station data from DWH via jretrieve
* implement configurable lapse-rate correction (default to true)

The lapse-rate correction is as follows:

#### Verification against KENDA-CH1
Baseline | Δz min (m) | Δz max (m) | Δz mean (m) | T_2M corr. min (K) | T_2M corr. max (K) | T_2M corr. mean (K)
-- | -- | -- | -- | -- | -- | --
baseline-3f32 (INCA) | −2877.7 | +2817.7 | −106.5 | −18.705 | +18.315 | −0.692
baseline-5ed2 (ICON-CH2-EPS) | −1349.6 | +813.9 | 0.0 | −8.773 | +5.290 | 0.000
baseline-b7ef (ICON-CH1-EPS) | 0.0 | 0.0 | 0.0 | — | — | —
baseline-b7ef (ICON-CH1-EPS) | 0.0 | 0.0 | 0.0 | — | — | —
cb7c (Varda-Single) | 0.0 | 0.0 | 0.0 | — | — | —

The large altitude differences when verifying INCA against KENDA-CH1 are due to the limited domain of INCA and the effect of nearest neighbour interpolation as illustrated in the three panels below. 

altitude_comparison_inca_kenda

#### Verification against SwissMetNet
Run | Δz min (m) | Δz max (m) | Δz mean (m) | T_2M corr. min (K) | T_2M corr. max (K) | T_2M corr. mean (K)
-- | -- | -- | -- | -- | -- | --
baseline-3f32 (INCA) | −606.4 | +705.2 | −33.1 | −3.942 | +4.583 | −0.215
baseline-5ed2 (ICON-CH2-EPS) | −745.9 | +736.1 | −78.4 | −4.848 | +4.785 | −0.509
baseline-b7ef (ICON-CH1-EPS) | −505.9 | +462.8 | −24.4 | −3.288 | +3.008 | −0.159
Varda-Single (cb7c) | −505.9 | +462.8 | −24.4 | −3.288 | +3.008 | −0.159

#### Lapse-rate corrected meteogram: 
202503030000_T_2M_JUN_lapse_rate

#### Meteogram without lapse-rate correction:
202503030000_T_2M_JUN_no_lapse_rate


jonasbhend:
There is yet another strange issue with altitude correction / topography. The diagnosis of our helpful friend below hints at the issue, but where it is coming from completely evades me. Can you help @clairemerker?

<html>
<body>
<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">When ICON-CH1-CTRL (or Varda-single) is compared to KENDA-CH1, the elevation coordinate for each source comes from a different place:</p>
Source | Elevation from
-- | --
Forecast (ICON-CH1-CTRL, Varda-single) | topography_c from external_parameter_icon_grid_0001_R19B08_mch.nc
Truth (KENDA-CH1 zarr) | FIS / 9.80665 stored in the zarr itself

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">These two fields are <strong>not the same</strong>. The comparison shows:</p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">dz = FIS/g − topography_c:
  min=−284.5 m, max=+285.4 m, mean=−0.0 m, std=6 m
  83 cells with |dz| &gt; 100 m (all in Alpine terrain)
</code></pre></div><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">This matches the log line exactly: <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">Lapse-rate correction: Δz range [-284.5, 285.4] m, mean -0.0 m</code>.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The 83 worst cells are all mountain/valley pairs in the Alps with alternating sign (e.g., cells 293370/293371 at lat≈45°N lon≈6°E have dz = +117 / −124 m). This is the classic pattern of two orographies with <strong>different smoothing</strong> applied: one orography raises a peak while the neighbor valley is lowered, the other does the opposite. The zarr's FIS was almost certainly generated from a different version of the ICON-CH1 external parameter file than the one currently used for operational ICON-CH1-CTRL.</p>
</body>
</html>

jonasbhend:
Here is some more analysis on the above issue. The grid altitude from KENDA-CH1 and ICON-CH1-EPS look visually identical (obviously), but the grid altitudes are not identical in the mountains and there is sometimes large differences in individual cells (up to almost 300m).

image

image

src/verification/__init__.py
```diff
@@ -47,7 +123,7 @@ def __init__(
         regions = {}
         # add inner region for ML evaluation
         regions["all"] = [
-            Polygon(list(zip([1.5, 16, 16, 1.5, 1.5], [43, 43, 49.5, 49.5, 43])))
+            Polygon(list(zip([3.17, 11.95, 11.95, 3.17, 3.17], [43.69, 43.69, 49.39, 49.39, 43.69])))
```
cosunae: could you document this magic numbers?

src/data_input/__init__.py
```diff
@@ -33,6 +34,15 @@
 
 ZERO_KELVIN = -273.15  # °C
 
+GRAVITY = 9.80665  # m/s² — standard gravity used to convert FIS geopotential to metres
```
cosunae: we could try to promote using earthkit constants: 
https://github.com/ecmwf/earthkit-meteo/blob/1d907be09ca616c76b73130ec988d3e2d699bcf5/src/earthkit/meteo/constants/constants.py#L56

to avoid in general having them defined in various packages with different precisions.

src/data_input/__init__.py
```diff
@@ -33,6 +34,15 @@
...
+
+ICON_CH1_CONST_GRIB = Path(
+    "/users/oprusers/osm/opr.inn/data/ICON_INPUT/ICON-CH1-EPS/lfff00000000c"
```
cosunae: maybe a suggestion here. This sounds like an internal path (which might change anytime by operations, see the reference to "inn" release). Also there might be the usual issues with permissions that some users are in the different group and not able to read this file. And last, we are locking it to run evalml where this home is mounted. 
Alternatively, we could 
curl -X GET https://data.geo.admin.ch/api/stac/v1/collections/ch.meteoschweiz.ogd-forecasting-icon-ch1/assets
https://opendatadocs.meteoswiss.ch/e-forecast-data/e2-e3-numerical-weather-forecasting-model?download-options=restapi#available-parameters

in a cache directory (we only need to wget if the file is not present in the cache)

cosunae:
Isnt the difference in height between KENDA and ICON_CH1 the same issue of the 16 bits grib rounding found recently?

jonasbhend:
> Isnt the difference in height between KENDA and ICON_CH1 the same issue of the 16 bits grib rounding found recently?

No, this difference resulted from the fact, that I was using the external_parameters file that is an input for ICON. According to Lukas Jansing, external parameters get modified internally to produce the actual topography. I have now switched to using the topography constants file directly and with this, there is no lapse-rate correction when verifying ICON-CH1-EPS/CTRL or Varda-Single against KENDA-CH1 or REA-L-CH1.

src/verification/__init__.py
jonasbhend: i have added a comment in the code, if you want me to document this somewhere else, please let me know @cosunae

src/data_input/__init__.py
jonasbhend: Went all the way and am using `earhtkit.meteo.vertical.geopotential_height_from_geopotential`.

src/data_input/__init__.py
jonasbhend: done

src/data_input/__init__.py
```diff
@@ -126,6 +232,16 @@ def load_analysis_data_from_zarr(
...
         ds = ds.rename({"cell": "values"})
 
+    # Derive elevation from FIS (surface geopotential, m²/s²) and assign as coordinate.
+    # FIS is constant in time, so drop the time dimension to get a purely spatial coord.
+    if "FIS" in ds:
+        elevation = ekdv.geopotential_height_from_geopotential(
+            ds["FIS"].isel(time=0, drop=True)
+        )
+        ds = ds.assign_coords(elevation=elevation)
+        if "FIS" not in params:
```
frazane: Are we ever going to get "FIS" in `params`, given that it's a forcing? If not, we can get rid of this and above we can also `ds.pop` when assigning to `elevation`.

src/data_input/__init__.py
```diff
@@ -869,14 +997,35 @@ def load_icon_baseline_from_grib(
...
+            topo = _load_icon_topography(const_grib)
+            if result.sizes["values"] == len(topo):
+                result = result.assign_coords(elevation=("values", topo))
+            else:
+                LOG.warning(
+                    "elevation not assigned: values=%d but %s has %d cells",
+                    result.sizes["values"],
+                    const_grib.name,
+                    len(topo),
+                )
```
frazane: What would be the cause of this? If there are known failure cases, can we explain them in the warning?

tests/unit/test_verification.py
```diff
@@ -97,3 +97,67 @@ def test_aggregate_results_n_samples_by_season_and_init_hour():
...
     assert int(out["n_samples"].sel(season="DJF", init_hour=12)) == 1
     assert int(out["n_samples"].sel(season="JJA", init_hour=0)) == 1
+
+
+# ---------------------------------------------------------------------------
+# apply_lapse_rate_correction
+# ---------------------------------------------------------------------------
+
+
+def _make_lapse_rate_datasets(fcst_elev, obs_elev, t2m=280.0, td2m=270.0):
```
frazane: Can we make this a fixture?

src/evalml/config.py
```diff
@@ -534,6 +534,10 @@ class ConfigModel(BaseModel):
         description="List of experiment participants, including forecaster/temporal downscaler ML runs and baselines.",
     )
     truth: TruthConfig | None
+    lapse_rate_correction: bool = Field(
```
frazane, line 537: I wonder if we could move this under `truth` schema? I find it makes more sense, but no strong opinion on this.

src/verification/__init__.py
```diff
@@ -21,6 +21,82 @@
...
+        )
+    else:
+        LOG.info(
+            "Lapse-rate correction: Δz range [%.1f, %.1f] m, mean %.1f m.",
+            float(dz_vals.min()),
+            float(dz_vals.max()),
+            float(dz_vals.mean()),
+        )
+
+    fcst = fcst.copy()
```
frazane: Why do we need `.copy()`?

frazane (commented):
Looks good already overall, thanks @jonasbhend! I left minor comments, and in the meantime I am running the varda-single config with the current code. Will let you know if I encounter any issue.

src/verification/__init__.py
```diff
@@ -46,8 +122,16 @@ def __init__(
...
         regions = {}
         # add inner region for ML evaluation
+        # this is the extent of the largest lat/lon box that is fully within the radar/INCA domain
         regions["all"] = [
-            Polygon(list(zip([1.5, 16, 16, 1.5, 1.5], [43, 43, 49.5, 49.5, 43])))
+            Polygon(
+                list(
+                    zip(
+                        [3.17, 11.95, 11.95, 3.17, 3.17],
+                        [43.69, 43.69, 49.39, 49.39, 43.69],
```
jonasbhend: We shouldn't change the default just yet. Discussions on how to handle this are ongoing.

src/verification/__init__.py
jonasbhend: to avoid side effects of functions. Now modifying inplace and making it more explicit through function name

src/data_input/__init__.py
jonasbhend: done

src/data_input/__init__.py
jonasbhend: I don't understand your comment. This is just to safeguard against non-matching grid information. Or are you arguing that we could just try to assign the coordinate and rely on xarray raising an error if dimensions do not match?

src/evalml/config.py
jonasbhend: No I prefer to leave it as is (even though it is a bit isolated here, I agree). To me the lapse-rate correction is distinctively NOT an attribute of the truth data, but of the forecast. So if anything we could add it to the runs section, but this would only make sense if we are planning to support the use case that we do lapse-rate correction for some of the forecast sources only (which I don't see happening for now).

tests/unit/test_verification.py
jonasbhend: done

src/data_input/__init__.py
frazane: Just trying to understand if there are known situations where we might get arrays of different sizes, in which case a warning like this makes sense but I would add an explanation of the likely cause. Otherwise if there are no known possible failures I would just let xarray raise an error, yes.

clairemerker (approved):

src/data_input/__init__.py
jonasbhend: done

PR #192, opened 2026-06-19
Add meteogram and centralise plot style

clairemerker:
AI-assisted implementation!

This PR adds meteograms for publication, centralises some ploting style things, uses long names for variables and tries out some colors :) 

Any feedback and changes welcome!

To run only plotting: 
```
evalml publication config/Varda-Single_paper.yaml -- \
  --allowed-rules publication_all publication_meteogram publication_figures \
                  data_download_eckit_geo_grids \
  --forcerun publication_meteogram publication_figures
``` 

Current version of the plots:

image
image

### DIFF
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -377,12 +377,12 @@ class Stratification(BaseModel):
     """Stratification settings for the analysis."""
 
     regions: List[str] = Field(
-        ...,
-        description="List of region names for stratification.",
+        default_factory=list,
+        description="List of region names for stratification. Empty list means no spatial stratification.",
     )
-    root: str = Field(
-        ...,
-        description="Root directory where the region shapefiles are stored.",
+    root: Optional[str] = Field(
+        None,
+        description="Root directory where the region shapefiles are stored. Required when regions is non-empty.",
     )
 
 
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -326,12 +326,35 @@ class ShowcaseConfig(BaseModel):
     )
 
 
+class PublicationMeteogramConfig(BaseModel):
+    """Case selection for the publication meteogram figure."""
+
+    init_time: str = Field(
+        ...,
+        description="Initialisation time (YYYYMMDDHHMM) of the case to plot.",
+    )
+    station: str = Field(
+        ...,
+        description="PeakWeather station ID to plot (e.g. KLO, GVE, LUG).",
+    )
+    params: List[str] = Field(
+        default=["T_2M", "TOT_PREC", "SP_10M", "DD_10M"],
+        description="Display parameters (one panel each); SP_10M/DD_10M are derived.",
+    )
+
+    model_config = {"extra": "forbid"}
+
+
 class PublicationConfig(BaseModel):
     """Configuration for the publication workflow."""
 
     enabled: bool = Field(
         default=False,
         description="Whether to generate publication figures.",
+    )
+    meteogram: Optional[PublicationMeteogramConfig] = Field(
+        default=None,
+        description="Publication meteogram case selection (omit to skip it).",
     )
 
 
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -227,11 +227,3 @@ rule verification_metrics_plot_all:
         ),
 
 
-rule publication_all:
-    """Target rule for publication figures workflow."""
-    input:
-        (
-            rules.publication_figures.output
-            if config.get("publication", {}).get("enabled", False)
-            else []
-        ),
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -260,3 +260,13 @@ rule verification_metrics_plot_all:
         ),
 
 
+rule publication_all:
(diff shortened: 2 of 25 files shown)


src/evalml/config.py
jonasbhend, line 1: Can't we reuse the existing MeteogramConfig class?

workflow/scripts/meteogram_derivations.py
jonasbhend, line 1: I don't understand. Why do we need this on top of the meteogram functionality that is already available in the showcases module / part? Seems like an unnecessary duplication to me.

workflow/scripts/meteogram_derivations.py
```diff
@@ -0,0 +1,76 @@
...
+
+def add_derived(ds: xr.Dataset, display_params: list[str]) -> xr.Dataset:
+    """Add requested derived variables (SP_10M, DD_10M) to a dataset that holds
+    U_10M/V_10M. Base params are left untouched."""
+    ds = ds.copy()
+    have_uv = "U_10M" in ds and "V_10M" in ds
+    if "SP_10M" in display_params and have_uv:
+        ds["SP_10M"] = wind_speed(ds["U_10M"], ds["V_10M"])
+    if "DD_10M" in display_params and have_uv:
+        ds["DD_10M"] = wind_direction_deg(ds["U_10M"], ds["V_10M"])
```
jonasbhend: Should an error be raised if something we cannot compute wind speed and direction (but would like to)?

workflow/scripts/meteogram_derivations.py
```diff
@@ -0,0 +1,76 @@
...
+    return ds
+
+
+def station_timeseries_to_long(
+    ds: xr.Dataset, source: str, display_params: list[str]
+) -> pd.DataFrame:
+    """Flatten a single-station dataset to long form.
+
+    Returns columns [source, valid_time, param, value]. The time coordinate is
+    `valid_time` if present, else `time`. Params absent from `ds` are skipped.
```
jonasbhend: Here again, should we raise an error if absent?

workflow/scripts/publication_style.py
jonasbhend, line 1: I like the idea behind this. unfortunately I have only now discovered the matplotlib style sheets (https://matplotlib.org/stable/users/explain/customizing.html), that probably would be more suitable, what do you think?

workflow/Snakefile
```diff
@@ -227,11 +227,3 @@ rule verification_metrics_plot_all:
...
 
 
-rule publication_all:
-    """Target rule for publication figures workflow."""
-    input:
-        (
-            rules.publication_figures.output
-            if config.get("publication", {}).get("enabled", False)
-            else []
-        ),
```
jonasbhend, line 270: All the other integral rules are in the top-level snakefile (e.g. experiment_all, ...). Is it necessary to move this one into the specific rule file?

jonasbhend (commented):
Thanks for pushing this. Really like the suggested colours. I think we should adopt the matplotlib style sheets for better readability.

workflow/rules/publication.smk
```diff
@@ -24,4 +24,74 @@ rule publication_figures:
...
+                f"{cfg['root']}|{cfg['steps']}|{cfg['member']}|{cfg['label']}"
+            )
+    return ";".join(specs)
+
+
+rule publication_meteogram:
+    input:
+        script="workflow/scripts/publication_meteogram.py",
+        peakweather_dir=rules.data_download_obs_from_peakweather.output.root,
+        eckit_grids=rules.data_download_eckit_geo_grids.output,
```
jonasbhend: Unfortunately, we already merged the jretrieve PR and thus have to refactor this part for the publications.

clairemerker: That's great actually, no worries :) I will refactor

workflow/scripts/publication_style.py
clairemerker: thanks for the PR!

workflow/scripts/meteogram_derivations.py
clairemerker: Thanks for catching that! Raising ValueError now.

workflow/scripts/meteogram_derivations.py
clairemerker: Thanks for catching that! Raising KeyError now

workflow/Snakefile
clairemerker: Good point, it's moved up again

src/evalml/config.py
clairemerker: I am happy to discuss to reduce duplication @jonasbhend ! But I would need some input, since I don't know how best to achieve that. The main reason to implement `PublicationMeteogramConfig` is that it allows to select one init time, `MeteogramConfig` loops over multiple, and I don't really understand how to separate the current `MeteogramConfig` from the whole showcase workflow since they share the `params`... 
As for the duplications of the meteogram functionalities, I suggest extending `meteogram_derivations.py` to also compute SP (pressure level wind speed) and reuse that in `plot_meteogram.py` for the showcase.

PR #193, opened 2026-06-19
Add forecaster support overlay to meteograms

adestefani8:
Adds an option to show the forecaster support points as dots on the meteograms.

Off by default, turn it on with:

```yaml
meteograms:
  forecaster_support: true
```

Only runs that have a forecaster are affected; the rest ignore the flag.

**Legend label:** the dots are labelled `forecaster support` by default, or with whatever `label` is set on the run's forecaster, if present.

Forecaster_Support_Demo_202506271800_U_10M_SIO


jonasbhend:
needs functionality from #247 to work in the new way of handling aggregations explicitly. Can be merged after #247

PR #194, opened 2026-06-19
Performance improvement in verification metrics computation

jonasbhend:
Due to my poor implementation choices, verification_metrics rules are slower than they need be. This PR provides a 5-6x (2x) speedup on analysis-based (station-based) computation of verification metrics. The improvement is achieved by the following changes:

1. Vectorise region loop (10% improvement, 294→266s)
Removed the inner for region loop by broadcasting masks as a leading dimension. Smaller dask graphs, less Python loop overhead. Modest gain on its own.

2. Rechunk forecast after map_forecast_to_truth (29% improvement, 266→189s)
isel() with fancy indices collapsed the forecast into one monolithic dask chunk. Adding fcst.chunk({"step": 1}) split it into 21 independent tasks, letting dask parallelise the time dimension for both fcst and obs simultaneously. Continuous scores dropped from ~10s to near-instant.

3. Vectorise contingency table (73% improvement, 189→52s)
Replaced the scores.categorical per-threshold loop with a single broadcast over a threshold dimension. Instead of N separate spatial passes (one per threshold), all thresholds are computed in one dask graph traversal. This was the dominant cost — ~50s per parameter — and is now negligible.

### DIFF
--- a/src/data_input/__init__.py
+++ b/src/data_input/__init__.py
@@ -392,9 +392,9 @@ def load_truth_data(
             params=params,
         )
         truth = truth.compute().chunk(
-            {"y": -1, "x": -1}
+            {"time": 1, "y": -1, "x": -1}
             if "y" in truth.dims and "x" in truth.dims
-            else {"values": -1}
+            else {"time": 1, "values": -1}
         )
     elif "peakweather" in str(root):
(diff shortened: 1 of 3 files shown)


frazane (approved):
Very nice, thanks!

PR #196, opened 2026-06-22
adopt style sheets

jonasbhend:
suggestion to address my reviewer comment

One visual change worth noting: xlabel "Lead time (h)" goes from 13 pt to 16 pt, matching the new axes.labelsize: 16. If you prefer it smaller (to match tick labels), set axes.labelsize: 13 in the stylesheet and the ylabel row-header sizing will drop accordingly.


clairemerker:
That's great, thanks!

PR #198, opened 2026-06-22
Remove labels from intermediate verification results

jonasbhend:
If evalml experiments are rerun with changed labels, this causes issues in the aggregated verification results as identification of scores and metrics is by labels (the dimension source in the verif_....nc files). To avoid potential confusion, we need to switch from human-readable labels to hash-based ids as elsewhere in evalml.

### Summary of changes
* Remove labels as source identifiers
* provide id:label key-value arguments to reports scripts for decoding ids


frazane (approved):
This makes sense, as a general rule of thumb labels should really just be for final deliverables. Looks good to go (if you can confirm it runs).

PR #199, opened 2026-06-22
Paper figures

jonasbhend:
WIP on figures used for Varda-Single publication


jonasbhend:
https://github.com/MeteoSwiss/evalml/pull/199/changes/d8409f91790049176897ea8114b94445307e44c8 provides a suggestion for an updated evaluation with the wind_speed and explicit aggregations from #203 . If you need to run before this is merged, please reset to previous stage.

PR #200, opened 2026-06-24
Fix meteogram obs for non-default stations and missing time import

clairemerker:
AI assisted implementation!

- Derive the jretrievedwh obs marker from --station inside the script instead of passing a separate --obs argument. The marker value contains '=' (e.g. jretrievedwh:locations=KLO), which marimo's cli_args drops, so obs silently fell back to the KLO default and any non-KLO station was broken (also via the snakemake rule). The rule no longer passes --obs.
- Import `time` in the entry cell and thread it through: the timing log lines used time.perf_counter() but `time` was never imported (NameError at runtime).


jonasbhend (approved):
Thanks for spotting this. Hard to catch if all you do is run the default configs :-/

PR #202, opened 2026-06-26
Refactor/figures

clairemerker:
AI assisted implementation! 

This PR is an attempt to improve the plotting workflow for publication to be able to generate them from the snakemake workflow as well as interactively. 

A documentation is currently available in `docs/publication_figures.md`

I only tested creating figures using both workflows, never triggered the whole pipeline incl. results creation... So not sure if the whole workflow is still ok, but I think it should...

### DIFF
--- a/config/varda-single_paper.yaml
+++ b/config/varda-single_paper.yaml
@@ -30,22 +30,22 @@ runs:
       root: /store_new/mch/msopr/osm/ICON-CH2-EPS
       steps: 0/120/1
       member: control
-  # - baseline:
-  #     label: ICON-CH1-EPS mean
-  #     root: /store_new/mch/msopr/osm/ICON-CH1-EPS
-  #     steps: 0/33/1
-  #     member: mean
-  # - baseline:
-  #     label: ICON-CH2-EPS mean
-  #     root: /store_new/mch/msopr/osm/ICON-CH2-EPS
-  #     steps: 0/120/1
-  #     member: mean
+  - baseline:
+      label: ICON-CH1-EPS mean
+      root: /store_new/mch/msopr/osm/ICON-CH1-EPS
+      steps: 0/33/1
+      member: mean
+  - baseline:
+      label: ICON-CH2-EPS mean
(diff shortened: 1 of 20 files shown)


jonasbhend (commented):
Looking really good. Thanks for seeing this through. One question. The current workflow still overwrites results based on stations and analysis, as both the manifest and the output paths are truth-hash-free (or experiment-hash-free). I think we should change that, such that we can run station and analysis-based workflows, without overwriting?

clairemerker:
@jonasbhend Good point! 

Do you mean the snakemake workflow only or also when runing the standalone plotting? 
For the snakemake workflow I don't see an issue, I can make the manifest and the ouput folder hash-dependent.
But then the standalone workflow needs to find the right manifest without knowing the hash... Not sure how to deal with that... Any idea?

And generally, is it enough to use the truth-hash, or do we need to hash the config?

jonasbhend:
> @jonasbhend Good point!
> 
> Do you mean the snakemake workflow only or also when runing the standalone plotting? For the snakemake workflow I don't see an issue, I can make the manifest and the ouput folder hash-dependent. But then the standalone workflow needs to find the right manifest without knowing the hash... Not sure how to deal with that... Any idea?
> 
> And generally, is it enough to use the truth-hash, or do we need to hash the config?

Well, we've done this before and it didn't work. But I would probably just use the truth label to denote the manifests. That way it is easy to operate in standalone mode and 'reasonably' fail-safe. If you trigger a full rerun, this will anyway be recomputed if the config changes and we probably don't want to keep old manifests lying around.

clairemerker:
@jonasbhend like this maybe? See last commit :)

jonasbhend (approved):
Hi @clairemerker. 
Das sieht super aus, vielen Dank. Ich habe es noch nicht ausprobiert, und komme heute erst nach dem weekly dazu. Von mir aus kannst du aber auch gerne schon in den paper-figures branch mergen, wenn es für dich funktioniert.

PR #203, opened 2026-06-30
Wind speed and explicit aggregations

jonasbhend:
This PR implements both derived parameters (e.g. wind speed from eastward and northward components) and explicit handling of aggregation periods. The latter allows us to evaluate hourly, 6-hourly and daily precipitation in the same evalml experiment or meteogram run. 

### Out of scope
Adopt the novel funtionality for plotting of animations of forecast fields. This is a somewhat larger refactor due to the different data layout and combination of high-resolution inner and global outer domain necessary for forecast maps. 

### Summary of changes
* Handle derived parameters in data_input
* Explicit aggregations for precipitation

### Example output

| TOT_PREC1 | TOT_PREC6 |
|---------------|-----------------|
| 202503100000_TOT_PREC1_KLO | 202503100000_TOT_PREC6_KLO |

image
Example dashboard: file:///M:/zue-prod/fc_development/seamless/S-RUC/evaluation/MRB-855_dashboard.html


dnerini:
Very cool, thanks for this much needed feature! The PR looks already very mature, I just wonder whether we could already use it to integrate over plain variables too? Eg 24h min/max temperature? Or would you say this is out of scope?

jonasbhend:
> Very cool, thanks for this much needed feature! The PR looks already very mature, I just wonder whether we could already use it to integrate over plain variables too? Eg 24h min/max temperature? Or would you say this is out of scope?

Good suggestion, I will have a look. Two things though:
1. daily minimum and maximum temperatures are not the same as daily minimum and maximum of the hourly instantaneous temperatures. So in essence we should at least think about modelling these separately in the long term.
2. The current aggregation is rolling, i.e. moving window. This is a bit counter to the daily min/max application, but maybe it is not too hard to integrate, I just don't know yet.

dnerini:
Good point! So maybe a better scope for this PR could be to stick to a rolling type of aggregation, eg 24h min/max/mean temperature?

jonasbhend:
That we could definitely do, yes. And if you select the 00 UTC timestamps, you basically have daily min/max

jonasbhend:
I was to quick there. No I don't think adding maximum  / minimum temperature is a good idea as we do not support different aggregating functions. The current PR works for sunshine duration, precipitation, etc., but I would suggest we can discuss and implement alternative aggregation functions in a separate PR. It's a good point though, as we will need to support alternative aggregation functions for wind gusts, temperature, etc. Moving window aggregation, however, is not very efficient. Here it works because for sums, you can just use the cumsum / lagged-diff trick.

src/data_input/__init__.py
```diff
@@ -37,6 +38,94 @@
...
+    if m and m.group(1) in _ACCUMULATABLE_PARAMS:
+        return m.group(1), int(m.group(2))
+    return param, None
+
+
+def compute_derived(ds: xr.Dataset, param: str) -> xr.DataArray:
+    """Compute a spatial derived variable from its components already in ds."""
+    if param == "SP_10M":
+        da = (ds["U_10M"] ** 2 + ds["V_10M"] ** 2) ** 0.5
+        # da.attrs["parameter"] = {
```
clairemerker: Are those attributes not needed anymore? I was thinking about the meteograms for example (https://github.com/MeteoSwiss/evalml/blob/cafea74f8039c81b01837e26c2069d8d1a8f08a7/workflow/scripts/plot_meteogram.py#L168)

src/data_input/__init__.py
```diff
@@ -1017,29 +1204,96 @@ def load_icon_baseline_from_grib(
...
+        )
+        ds = _load_INCA_baseline_from_netcdf(
+            root, reftime, load_steps, inca_load_params
+        )
+    else:
+        LOG.info("Loading baseline forecasts from ICON GRIB archive...")
+        ds = _load_icon_baseline_from_grib(
+            root, reftime, load_steps, load_params, member=member
+        )
+    LOG.info(ds)
```
clairemerker: Drop or make `LOG.debug(ds)`?

workflow/scripts/verification_scoremaps.py
```diff
@@ -443,14 +228,9 @@ def main(args: Namespace) -> None:
...
         # --- load truth slice ---
-        truth_slice = truth_da.sel(time=valid_time).compute()
-        # For derived variables truth_da is already the derived DataArray,
-        # so wrap it in a Dataset for map_forecast_to_truth compatibility.
-        truth_ds = (
-            truth_slice.to_dataset(name=args.param)
-            if isinstance(truth_slice, xr.DataArray)
-            else truth_slice
-        )
+        truth_ds = load_truth_data(args.truth, reftime, [args.step], [args.param])
```
clairemerker: I might be wrong, but isn't this opening the zarr at every iteration? I think the workflow in main opened it once (lazy) before the `for reftime` loop. Could that be an issue when aggregating over a long period?

workflow/scripts/verification_scoremaps.py
```diff
@@ -331,23 +158,6 @@ def main(args: Namespace) -> None:
...
 
-    # Every configured init must have a matching truth slice; a gap here would
-    # otherwise silently drop the init from the map. Check up front so the full
-    # set of missing valid times is reported at once rather than one per run.
-    required_valid_times = {
-        np.datetime64(rt + step_td).astype("datetime64[ns]") for rt, _ in init_items
-    }
-    missing_truth = sorted(required_valid_times - truth_times)
-    if missing_truth:
-        raise ValueError(
```
clairemerker, line 342: Do we now possibly fail much later in the loop if there is a mismatch, instead of here?

src/data_input/__init__.py
```diff
@@ -349,97 +462,93 @@ def fieldlist_to_xarray(fieldlist) -> xr.Dataset:
...
-        raise ValueError(
-            "TOT_PREC in the GRIB appears to already be "
-            f"period-accumulated (min(.diff()) = {min_diff:.3e} m). "
-            "Check that the accumulate_from_start_of_forecast post-processor "
-            "is enabled in the anemoi-inference config for this source."
-        )
 
-    # Clip remaining small negative values to zero
-    tp = tp.clip(min=0.0)
+def _accumulate_from_hourly(da_1h: xr.DataArray, steps: list[int]) -> xr.DataArray:
```
clairemerker, line 548: I would suggest adding at least a round-trip test for `_accumulate_from_hourly` → `_disaggregate_accum`, and a check for negative value?

clairemerker (approved):
AI assisted review ;) 

I added a few comments, feel free to address or not!

src/data_input/__init__.py
jonasbhend: done

workflow/scripts/verification_scoremaps.py
jonasbhend: Nice catch 🎣. Claudia suggested a fix but I haven't tested this yet. Hope this works (at least the issue is addressed from what I can tell).

workflow/scripts/verification_scoremaps.py
jonasbhend: again 🎣. Fixed

src/data_input/__init__.py
jonasbhend: round-trip test added in https://github.com/MeteoSwiss/evalml/pull/203/changes/d017f913123e1b016749450d53aacb665459087a

src/data_input/__init__.py
jonasbhend: yep, true. The issue is that these attributes are not serializeable by xarray.Dataset.to_netcdf. So as an intermediate fix, the attributes are removed before serialization, thus keeping them available for meteogram plotting. Implemented in https://github.com/MeteoSwiss/evalml/pull/203/changes/f5f24b52db4bdc1d11a0a7acc0477fcd07f5228c

PR #204, opened 2026-06-30
Support global models

dnerini:
supersedes #90 following #190

additionally:
- add support for cloud cover and surface radiation variables
- add support for rotating animations


src/verification/__init__.py
```diff
@@ -45,24 +45,33 @@ def __init__(
         ).transform
 
         regions = {}
-        # add inner region for ML evaluation
-        regions["all"] = [
-            Polygon(list(zip([1.5, 16, 16, 1.5, 1.5], [43, 43, 49.5, 49.5, 43])))
-        ]
-        if shp and shp != [""]:
+        has_shapefiles = bool(shp and shp != [""])
+        if has_shapefiles:
```
MicheleCattaneo: If I understand correctly, now as soon as there is a shapefile in the regions, there is a "hardcoded" `all` region that is the ICON domain. If no shapefile is passed, the behaviour changes for regional models, where the results are going to be global.

Potential issues I see:
- Would we ever pass other shapefiles that are not within our hardcoded `all`?
- We should make it clear that for regional forecaster, not passing any region shapefile changes the reulsts. 


What do you think?

dnerini: correct and I introduced it mostly for backward compatibility. But maybe we can try to find a better logic here? @jonasbhend is also refining the hardcoded domain as part of #191

MicheleCattaneo: But my point is that it's not backward compatible, or am I missing something? 
What was the expected domain of verification if no shapefile was passed? With this PR, it is a global verification. I thought that it was **always** regional so far.

src/data_input/__init__.py
```diff
@@ -349,6 +349,13 @@ def load_forecast_data_from_grib(
     if ifs_rename:
         ds = ds.rename(ifs_rename)
 
+    if "tp" in ifs_rename:
```
MicheleCattaneo: @dnerini could you confirm that this fix that I introduced is not breaking any logic for regional forecaster? Without this change, global truth data had a different unit and was the source of the TOT_PREC metrics being the same across multiple forecasters (the metrics were overlapping in the plot)

src/verification/__init__.py
dnerini: it's backward compatible to the extent that we usually include regions for stratification, but otherwise no, results will change for an empty lists of regions

jonasbhend: What is the intended behavior for global models? Verify on the globe?

Couldn't we check the list of regions againts the data domain and only iterate on those that are fully within the data domain? That way, we could use a different set of shapefiles for global and regional analysis. Not sure how to elegantly handle the different defaults for global (take all the data) and regional (take the inner 'all' domain) though. I guess with a simple conditional should work for now.

src/data_input/__init__.py
dnerini: Mmmh good point, I think now the right approach should be to use a postprocessor instead to scale the output for precip. I'll have a closer look!

src/data_input/__init__.py
```diff
@@ -126,6 +126,16 @@ def load_analysis_data_from_zarr(
...
+    # Drop grid points with undefined (NaN) coordinates
+    if "values" in ds.dims and "latitude" in ds.coords and "longitude" in ds.coords:
+        valid = np.isfinite(ds["latitude"].values) & np.isfinite(ds["longitude"].values)
+        if not valid.all():
+            LOG.warning(
+                "Dropping %d grid point(s) with undefined lat/lon from truth dataset.",
+                int((~valid).sum()),
+            )
+            ds = ds.isel(values=valid)
+
```
radiradev, line 349: Why would there be infinite values in coordinates?

src/data_input/__init__.py
dnerini: ec82cfd

src/data_input/__init__.py
dnerini: I looked a bit into it, turns out that the zarr arrays store fill_value=0.0, and xarray's open_zarr masks any value equal to the fill value as NaN, which occurs at 0° coordinates....

src/verification/__init__.py
dnerini: OK I have a new suggestion wrt verification regions and the behaviour of the default "all" region, the new configuration could look like this:
```
stratification:
  regions:
    - all: [1.5, 16, 43, 49.5]                # restrict default region
    - alpine_south: [7.5, 12, 43, 46]  # arbitrary named bbox
    - mittelland                                # shapefile as before
  root: /store_new/mch/msopr/ml/regions/...
```
where "all" is only special in that `ShapefileSpatialAggregationMasks` initializes it to None (ie full domain of the ground truth), while any other bbox dict entry just adds to regions after construction.

thoughts? @MicheleCattaneo @jonasbhend ?

dnerini: see acac66e

The `regions` list in `experiment.stratification` now accepts two entry types: plain strings (shapefile names, as before) and single-key dicts mapping a region name to a `[lon_min, lon_max, lat_min, lat_max]` bounding box. The special key all controls the default region: if omitted, it covers the full ground-truth domain (useful for global evaluation). All existing configs that used shapefiles now explicitly carry `- all: [1.5, 16, 43, 49.5]` (the previously hardcoded verification domain). The config model, schema, Snakemake rules, and the verification script are all updated consistently, with `shp_regions` and `bbox_regions` as the two distinct CLI arguments.

jonasbhend: Good suggestion to make this more explicity (and more easily configurable). Not sure I understand your proposition. Wouldn't something like this be more intuitive?

```
stratification:
  regions:
    - all                                # unrestricted complete domain
    - icon: [1.5, 16, 43, 49.5]          # ICON domain (minus some 'sponge' zone)
    - radar: [3.17, 11.95, 43.69, 49.39] # INCA domain
    - alpine_south: [7.5, 12, 43, 46]    # arbitrary named bbox
    - mittelland                         # shapefile as before
  root: /store_new/mch/msopr/ml/regions/...
 ```
 
 That way you can opt out of the 'redundant' all domain if you are evaluating high-res data only.

Comment (jonasbhend) [reply src/verification/__init__.py]: In addition, we should probably make sure that we don't extrapolate ad infinitum to avoid misinterpreting INCA results (i.e. we would get missing values for the icon domain for INCA which would immediately signal that we don't have full coverage).

Comment (jonasbhend) [reply src/verification/__init__.py]: or as in the above but with 
```
        - all: []                            # unrestricted complete domain
```
for even more clarity

Comment (dnerini) [reply src/verification/__init__.py]: it was mostly an attempt to provide a way to reproduce the same results as before, by manually specifying what "all" should be, but I'm happy to remove that logic if you don't see the reason to do so

File: config/stage-a-o96-multi-step.yaml [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+  params:
+    - T_2M
+    - SP_10M
+    - TOT_PREC
+    - CLCT
+  meteograms:
+    enabled: false # Because to Jretrieve credentials
+  animations:
+    enabled: true
+    domains:
```
dnerini: @MicheleCattaneo I think this new syntax from d308e5c is not supported by the config schema? it breaks the pydantic validation

src/verification/__init__.py
dnerini: and I assume we always want a "all" region, even if omitted from the list?

config/stage-a-o96-multi-step.yaml
MicheleCattaneo: I will check, thanks

src/verification/__init__.py
jonasbhend: Oh that is a slightly hard one. I would love for region names to be unique, i.e. the above proposal of 'all' just taking whatever is there is definitely not providing that. But the alternative of doing:

```
stratification:
  regions:
    - global: [-Inf, Inf, -Inf, Inf]     # unrestricted complete domain
    - icon: [1.5, 16, 43, 49.5]          # ICON domain (minus some 'sponge' zone)
    - radar: [3.17, 11.95, 43.69, 49.39] # INCA domain
    - alpine_south: [7.5, 12, 43, 46]    # arbitrary named bbox
    - mittelland                         # shapefile as before
  root: /store_new/mch/msopr/ml/regions/...
```
with some check on domain size and exclusion, or auto-adding icon or global depending on the size of the input data is also somewhat unsatisfactory. In a way I am leaning to the above and the user has to make sure that the selection of regions is appropriate.

jonasbhend: Minor updates:
1. #191 didn't change the extent of the inner verification domain
2. #212 is adding masking for incomplete domain coverage

MicheleCattaneo: Is the domain selection now as you want it? I think we are close to a merge. Let me know if there is still something to be done

jonasbhend: I would still argue that we should take this opportunity to harmonize the region definitions and make regions explicit. I.e. avoid the `all` region that will be the icon domain or global domain depending on the use case. Instead I propose to use:

1. default to `icon` domain if no stratification is provided
2. use `global` for the global verification to avoid confusion

This of course implies that we have to select the icon domain by default in the dashboard and plots.

MicheleCattaneo: What do you think about avoiding defaults and having the schema expect an explicit value for the region of interest?

jonasbhend: Here's another proposal:
1. we disallow region = None and make selection of regions explicit (i.e. we can use icon and global as names)
2. first region gets plotted and used in dashboard

src/data_input/__init__.py
```diff
@@ -25,6 +25,15 @@
     "2d": "TD_2M",
     "sp": "PS",
     "lsm": "FR_LAND",
+    "tcc": "CLCT",
+    "lcc": "CLCL",
```
jonasbhend, line 30: Why only total and low cloud cover? We could produce all three, right?

src/data_input/__init__.py
```diff
@@ -25,6 +25,15 @@
...
+    # TODO: ssrd is treated as a plain per-step field (no de-accumulation),
+    # which only holds because it's not currently listed in any
+    # accumulate_from_start_of_forecast.accumulations in the inference
+    # configs (unlike tp, see _tot_prec_handling). If ssrd/strd are ever
+    # added there, this needs the same cumulative-since-start handling tp
+    # gets, or verification/plots will silently be wrong.
```
jonasbhend, line 36: So we already produce this, but in fact it is accumulated (or not)? I am confused...

src/data_input/__init__.py
jonasbhend: But since this is deterministic and we don't expect missing coordinates for other reasons, wouldn't it be better to fix the root cause, i.e. something like:

```
ds["longitude"] = ds["longitude"].fillna(0.0)
ds["latitude"] = ds["latitude"].fillna(0.0)
```

src/verification/__init__.py
dnerini: commit 69099e2 implements your suggestion, can you have a look?

config/aifs-single.yaml
```diff
@@ -18,30 +18,31 @@ runs:
...
+
+lapse_rate_correction: false
 
 experiment:
   params:
     - T_2M
     - TD_2M
     - SP_10M
-    - TOT_PREC6
+    - TOT_PREC
```
jonasbhend: This shouldn't work with the current implementation of evalml. no?

config/forecasters-ich1-oper.yaml
```diff
@@ -44,6 +44,7 @@ experiment:
     - TOT_PREC6
   stratification:
     regions:
+      - icon-ch1: [1.5, 16, 43, 49.5]
```
jonasbhend: As stated before, I suggest to not use `icon-ch1` as the domain is independent of the resolution.

src/evalml/config.py
```diff
@@ -382,15 +402,45 @@ class Locations(BaseModel):
...
+    @field_validator("regions")
+    @classmethod
+    def validate_regions(
+        cls, v: List[Union[str, Dict[str, List[float]]]]
+    ) -> List[Union[str, Dict[str, List[float]]]]:
+        if not v:
+            raise ValueError(
+                "At least one region must be specified. "
+                "Add a domain bbox as the first entry, e.g. regions: [{global: [-180, 180, -90, 90]}] "
+                "for global models or [{icon-ch1: [1.5, 16, 43, 49.5]}] for ICON-CH1."
```
jonasbhend: The first region doesn't have to be a bboxed region, right? The comment seems to imply this.

src/verification/__init__.py
```diff
@@ -328,8 +337,12 @@ def verify(
         Label for the forecast source (used in output dataset).
     obs_label : str
         Label for the observation source (used in output dataset).
-    regions : list[str] or None, optional
-        List of shapefile paths or region names to use for spatial aggregation. If None, uses default region ('all').
+    regions : list[dict] or None, optional
+        Ordered list of region specs. Each entry is either
```
jonasbhend: i think this is wrong, no? Below you throw an error if regions is none, so none should not be supported.

config/forecasters-ich1.yaml
```diff
@@ -57,6 +57,7 @@ experiment:
     - PMSL
   stratification:
     regions:
+      - icon-ch1: [1.5, 16, 43, 49.5]
```
jonasbhend: Instead of using the bbox framing, shouldn't we use reserved keywords for pre-defined regions, e.g.:

```
icon # defaults to [1.5, 16, 43, 49.5]
global # defaults to [0, 360, -90, 90]
```

in order to prevent users from mislabelling standard regions?

jonasbhend (approved):
I have a few minor comments, otherwise looking good. Sorry for taking so long.

src/data_input/__init__.py
dnerini: this part is a bit out of scope to this PR, I suggest to leave it to a minumum

src/data_input/__init__.py
dnerini: @MicheleCattaneo ?

config/aifs-single.yaml
dnerini: uhm actually it didn't work with TOT_PREC6, I assumed because the era5 dataset doesn't include it, but maybe it's just a problem of mapping to the right variable naming. need to investigate this a bit more,

config/forecasters-ich1.yaml
dnerini: ok, but we still want to leave the option to pass arbitrary bboxes, right?

src/data_input/__init__.py
jonasbhend: just for my enlightenment, not really important.

config/forecasters-ich1.yaml
jonasbhend: yes, of course. But I was just thinking that for regions that we plan to reuse or set as default, a keyword approach would lead to less surprises and be quicker.

dnerini: another question: should we unify pre-defined regions and domains used for plotting? currently we would have `icon # defaults to [1.5, 16, 43, 49.5]` for verification, but `icon-ch # domain extent [0, 17.5, 40.5, 53.0]` for plotting

jonasbhend: I think the scope here is different. For verification we want to exclude the sponge zone (i.e. clip quite a bit to the inside of the domain), whereas for plotting we may indeed want to show the boundary (or at least we did so far).

dnerini: sure, for this we could just have two pre-defined domains, e.g (icon-verif, icon-ch), but still it remains the question of whether we should unify pre-defined regions bboxes and plotting domains. 

--> let's tackle this in a sperate PR (if indeed we decide to refactor it)

dnerini:
> I have a few minor comments, otherwise looking good. Sorry for taking so long.

thanks! I addressed your comments above, hopefully a final look?

src/data_input/__init__.py
dnerini: my understanding is that we predict the same variable used during training, which is period accumulated

MicheleCattaneo: Sorry missed the comment.
We produce it only on some global experiments are the moment. The unit is W/m^2, and Watts are depending on a time of accumulation, so it's similar to precipitation in that way, and it could be produced with the accumulation from the start of the forecast just like precip. For simplicity now I treated  is as an instantaneous variable (and the accumulation depends on the dataset used, 1h or 6h)

PR #205, opened 2026-07-01
Add fps setting for animation to config

lclanzi:
Add an optional `fps` (frame per second) field to `AnimationsConfig` to allow users to explicitly control the frame rate of forecast GIF animations. When not set, the previous behaviour is preserved, i.e. the delay is derived from the forecast step interval.

## Usage

To override the default frame rate, add `fps` under `showcase.animations` in the experiment config:

```yaml
showcase:
  animations:
    enabled: true
    fps: 0.2   # slow animation, one frame every 5 seconds

### DIFF
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -305,10 +305,13 @@ class AnimationsConfig(BaseModel):
             "[lon_min, lon_max, lat_min, lat_max], and optional 'projection'."
         ),
     )
-    fps: float = Field(
-        default=2.0,
+    fps: float | None = Field(
+        default=None,
         gt=0,
-        description="Frames per second for the output GIF animation.",
+        description=(
+            "Frames per second for the output GIF animation. "
+            "Defaults to one frame per forecast step interval."
+        ),
     )
 
 
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -305,6 +305,11 @@ class AnimationsConfig(BaseModel):
             "[lon_min, lon_max, lat_min, lat_max], and optional 'projection'."
         ),
     )
+    fps: float = Field(
+        default=2.0,
+        gt=0,
+        description="Frames per second for the output GIF animation.",
+    )
 
 
 class ScorecardConfig(BaseModel):
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -305,13 +305,10 @@ class AnimationsConfig(BaseModel):
             "[lon_min, lon_max, lat_min, lat_max], and optional 'projection'."
         ),
     )
-    fps: float | None = Field(
-        default=None,
+    frames_per_second: float = Field(
+        default=2.0,
         gt=0,
-        description=(
-            "Frames per second for the output GIF animation. "
-            "Defaults to one frame per forecast step interval."
-        ),
+        description="Frames per second for the output GIF animation.",
     )
 
 
--- a/workflow/rules/plot.smk
+++ b/workflow/rules/plot.smk
@@ -155,11 +155,7 @@ rule make_forecast_animation:
         region="|".join(map(re.escape, SHOWCASE_REGIONS.keys())),
     localrule: True
     params:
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: workflow/rules/plot.smk [hunk 1]
```diff
@@ -155,7 +155,11 @@ rule make_forecast_animation:
         region="|".join(map(re.escape, SHOWCASE_REGIONS.keys())),
     localrule: True
     params:
-        delay=lambda wc: 10 * int(RUN_CONFIGS[wc.run_id]["steps"].split("/")[2]),
+        delay=lambda wc: (
+            round(100 / SHOWCASE_ANIMATIONS_FPS)
+            if SHOWCASE_ANIMATIONS_FPS is not None
+            else 10 * int(RUN_CONFIGS[wc.run_id]["steps"].split("/")[2])
```
jonasbhend: Wouldn't it be easier to read, if we set a default for the config element? That way, we can always count on fps to be present. would be ok for me if this were to change existing behaviour (i.e. you have to specify it explicitly to get something meaningful in some circumstances).

jonasbhend (commented):
Hi @lclanzi Thanks for tackling this. I suggest to include the new config in one of the example configs for reference (even though it is optional).

src/evalml/config.py
```diff
@@ -305,6 +305,14 @@ class AnimationsConfig(BaseModel):
             "[lon_min, lon_max, lat_min, lat_max], and optional 'projection'."
         ),
     )
+    fps: float | None = Field(
```
jonasbhend: Also, this could be made easier to read by naming frames_per_second instead, what do you think?

lclanzi: Good idead, `fps` has been renamed to `frames_per_second`

workflow/rules/plot.smk
lclanzi: Yes, you are right. The default is now `frames_per_second=2`

workflow/rules/plot.smk
```diff
@@ -155,7 +155,7 @@ rule make_forecast_animation:
         region="|".join(map(re.escape, SHOWCASE_REGIONS.keys())),
     localrule: True
     params:
-        delay=lambda wc: 10 * int(RUN_CONFIGS[wc.run_id]["steps"].split("/")[2]),
+        delay=round(100 / SHOWCASE_ANIMATIONS_FPS),
```
dnerini: I don't think it's necessary to declare the global variable SHOWCASE_ANIMATIONS_FPS, we should be able to access it directly from RUN_CONFIGS, right? 

and do I understand correctly that you want to remove the previous logic that uses the steps parameter to derive the delay?

lclanzi: Yes, will look into that. 

I removed the previous default logic based on a suggestion from @jonasbhend (see above). But I can keep it if necessary

dnerini: ok let's keep the new default behavior then

lclanzi: > I don't think it's necessary to declare the global variable SHOWCASE_ANIMATIONS_FPS, we should be able to access it directly from RUN_CONFIGS, right?

I think that RUN_CONFIGS gathers information under the `run` block. However, `frame_per_secons` is under `showcase.animation`, which is never touched by `collect_all_runs()`. Therefore, it is not possible to define SHOWCASE_ANIMATIONS_FPS in RUN_CONFIGS.

dnerini: ah you're right! but then I would suggest to collect all the showcase config into one dictionary only, no?

lclanzi: all showcase animation params are now in one dict. I think we could also gather all params for showcase into SHOWCASE_CONFIGS, so that we would replicate the pattern adopted for RUN_CONFIGS. However, this would require quite some changes and I am not sure this PR is the right place to do so

dnerini (approved):

PR #206, opened 2026-07-01
Remove release candidate versions for earthkit tools

andreaspauling:
The stable versions of the earthkit packages can be used now

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,11 +26,11 @@ dependencies = [
     "scores>=2.0.0",
     "eccodes>=2.44,<2.48",
     "eccodes-cosmo-resources-python==2.44.0.1",
-    "earthkit-data==1.0.0rc10",
-    "earthkit-utils==1.0.0rc1",
-    "earthkit-plots==1.0.0rc9",
-    "earthkit-meteo==1.0.0rc3",
-    "earthkit-geo==1.0.0rc7"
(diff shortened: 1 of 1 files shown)


frazane:
Thanks for opening this PR! Does this also solve any pending issue related to grid coordinates?

andreaspauling:
> Thanks for opening this PR! Does this also solve any pending issue related to grid coordinates?

Unfortunately not. It just solved a compatibility issue after ` uv sync`. 

`verification_metrics_baseline` and `verification_metrics` jobs failed with:

eckit::codec::InvalidRecord: version not found in record
~/.local/share/eckit/geo/grid/icon/bbbd5a...ek

which is solved now.

frazane (changes requested):
I think we can expect ECMWF to publish patch releases (`1.0.X`) relatively often over the next few weeks. Perhaps it would be better not to pin a specific version but a range, e.g. 

```
earthkit-xxx>=1.0,<1.1
```

andreaspauling:
> I think we can expect ECMWF to publish patch releases (`1.0.X`) relatively often over the next few weeks. Perhaps it would be better not to pin a specific version but a range, e.g.
> 
> ```
> earthkit-xxx>=1.0,<1.1
> ```

That could make sense. Done.

frazane (approved):

PR #211, opened 2026-07-03
Fail fast if jretrieve credentials are missing

jonasbhend:
This PR implements checks on jretrieve credential to fail fast and with an informative error message if jretrieve credentials are missing and verificiation is performed against station data.

### Summary of changes
check jretrieve credentials as part of check_prerequisites function


clairemerker (approved):
Danke!

PR #212, opened 2026-07-06
Avoid extrapolation in nearest neighbour interpolation

jonasbhend:
When truth and forecast data are mapped, forecasts are extrapolated ad infinitum even if the forecast data covers only a subset of the domain. This is currently the case for INCA forecasts and the inner verification domain (aka most of the high-resolution domain). To avoid misinterpretation of evaluation results, we propose to suppress extrapolation and report missing values for regions that are not fully covered. This PR implements this change.

### Spatial coverage fraction
To allow for varying number of stations with observations by parameter, we compute the missing value fraction as the number of non-missing forecast / obs pairs divided by the number of non-missing observations. Currently the missing-value-fraction threshold is set to zero, because in this framework, we don't expect any missing forecasts other than incomplete domain coverage, which should indeed be masked.

### Summary of changes
* report distance between nearest neighbours
* estimate native grid spacing
* set extrapolated values to missing based on distnace with nearest neighbour and estimated grid spacing
* compute missing value fraction for area averages of verification results and mask results where too many points are missing
* do not mask observation metrics, as in the current implementation observation coverage is 'complete' by definition

### DIFF
--- a/src/verification/__init__.py
+++ b/src/verification/__init__.py
@@ -312,7 +312,7 @@ def verify(
     dim: list[str] | None = None,
     threshold_dict: dict[str, dict[str, list[float]]] | None = None,
     num_workers: int | None = None,
-    max_missing_fraction: float = 0.2,
+    max_missing_fraction: float = 0.0,
 ) -> xr.Dataset:
     """
     Compute verification metrics and statistics comparing forecast and observation datasets.
@@ -342,10 +342,11 @@ def verify(
     num_workers : int or None, optional
         Number of parallel workers for computation. If None, uses available CPU cores minus 2.
     max_missing_fraction : float, optional
-        Maximum allowed fraction of missing values (from either fcst or obs) within a region
+        Maximum allowed fraction of missing forecast values among obs-valid in-region points
         before a metric is set to NaN. Computed per region and time step over the reduction
-        dimensions. Default is 0.2 (20%). This allows legitimate station dropouts while
-        suppressing metrics where the forecast does not cover the region.
+        dimensions. Default is 0.0 — no missing forecasts are tolerated where observations
+        exist. Increase to a small positive value (e.g. 0.05) if spurious forecast gaps need
+        to be accommodated.
 
     Returns
     -------
@@ -397,11 +398,11 @@ def verify(
         fcst_param = fcst_aligned[param].where(masks)
         obs_param = obs_aligned[param].where(masks)
 
-        # Missing fraction: proportion of in-region points where either source is NaN.
-        # Computed with skipna=True so only in-region points (non-masked by region) count.
+        # Missing fraction: among obs-valid in-region points, fraction where fcst is missing.
+        # Normalising by obs availability avoids penalising parameters with fewer stations.
         missing_fraction = (
-            (fcst_param.isnull() | obs_param.isnull())
-            .where(masks)
+            fcst_param.isnull()
+            .where(obs_param.notnull() & masks)
             .mean(dim=dim, skipna=True)
         )
         too_many_missing = missing_fraction > max_missing_fraction
--- a/src/verification/__init__.py
+++ b/src/verification/__init__.py
@@ -420,16 +420,14 @@ def verify(
             prefix=param + ".",
             source=fcst_label,
             dim=dim,
-        )
+        ).where(~too_many_missing)
         obs_stats = _compute_statistics(
             obs_param,
             prefix=param + ".",
             source=obs_label,
             dim=dim,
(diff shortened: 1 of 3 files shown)


src/verification/spatial.py
```diff
@@ -119,6 +157,13 @@ def map_forecast_to_truth(fcst: xr.Dataset, truth: xr.Dataset) -> xr.Dataset:
     truth
         Reference dataset with `latitude` and `longitude` coordinates on either
         `(y, x)` or `values`.
+    extrapolate
+        If False (default), truth points whose nearest forecast point is farther
+        away than the estimated native grid spacing of the forecast are set to
+        missing in the returned dataset. The native spacing is the median
+        adjacent-cell chord distance for 2-D grids and the median
+        nearest-neighbour chord distance within the source for scattered points.
```
dnerini: this is just hypothetical, right? we don't cover the case of scattered forecast points (yet)

src/verification/spatial.py
```diff
@@ -103,7 +111,37 @@ def nearest_grid_yx_indices(
...
+        axis=-1,
+    )
+    if lat.ndim == 2:
+        dy = np.linalg.norm(xyz[1:] - xyz[:-1], axis=-1)
+        dx = np.linalg.norm(xyz[:, 1:] - xyz[:, :-1], axis=-1)
+        return float(np.median(np.concatenate([dy.ravel(), dx.ravel()])))
+    else:
+        pts = xyz.reshape(-1, 3)
+        tree = cKDTree(pts)
+        dists, _ = tree.query(pts, k=2)  # k=2 to skip the self-match (dist=0)
```
dnerini, line 138: do we need to consider cases with less than 2 points?

src/verification/__init__.py
```diff
@@ -389,32 +397,43 @@ def verify(
         fcst_param = fcst_aligned[param].where(masks)
         obs_param = obs_aligned[param].where(masks)
 
+        # Missing fraction: among obs-valid in-region points, fraction where fcst is missing.
+        # Normalising by obs availability avoids penalising parameters with fewer stations.
+        missing_fraction = (
+            fcst_param.isnull()
+            .where(obs_param.notnull() & masks)
```
dnerini: here `& masks` looks redundant given line 397 above?
```suggestion
            .where(obs_param.notnull())
```

src/verification/__init__.py
```diff
@@ -310,6 +312,7 @@ def verify(
     dim: list[str] | None = None,
     threshold_dict: dict[str, dict[str, list[float]]] | None = None,
     num_workers: int | None = None,
+    max_missing_fraction: float = 0.0,
```
dnerini, line 324: isn't this default too strict? I can imagine individual truth stations near borders being set to nan because too far from any forecast grid point (given `extrapolate=False`), but then the whole region being excluded for a single nan out of potentially hundreds of valid ones. Instead, this could add small tolerance in the default. This is even more important, since the argument is currently not exposed anywhere in the configs (essentially, the value is hardcoded)

dnerini (commented):
nice work! I left only a few minor comments

src/verification/__init__.py
jonasbhend: The problem here is that we want to guarantee that two competing forecasts are evaluated on exactly the same observations. This is only possible when setting max_missing_fraction to zero. The idea is that for forecasts that do not cover the whole domain, the value is missing (because we do not extrapolate), such that we do not inadvertently compare say INCA forecasts with ML/ICON forecasts on the whole ICON domain (which is not covered by INCA). For all sub-region stratifications (and in particular Switzerland and the forecasting regions) all the points are in the domain, so `max_missing_fraction=0.0` has no effect. I will add some more documentation to explain why this is set to 0 and not exposed via configs.

src/verification/spatial.py
jonasbhend: This is for flat horizontal coordinates, not cases with less than 2 points (or am I misunderstanding your comment)? Can you please resolve if you agree @dnerini ?

src/verification/spatial.py
jonasbhend: Again, this is an implementation detail, but for triangular grids such as ICON, the nearest-neighbour chord distance is used, whereas for regular grids, we can compute directly from distance in x/y.

Of course the approach for the ICON grid would technically also work on scattered forecast points (e.g. stations), but that is not what we are trying to do here.  I made this more explicit in the documentation of the function.

dnerini (approved):
Looking good to me (once tests pass)!

PR #213, opened 2026-07-10
Fix broken evalml runs introduced with MEC PR

jonasbhend:
#110 broke non-MEC evalml workflows and configs (my bad, sorry for not checking thoroughly). Here is a quick fix to rectify this.

### Summary of changes
1. `_reftimes_mec()`: replaced hardcoded `config["runs"][0]["forecaster"]["steps"]` with `RUN_CONFIGS` candidate lookup, so it works for temporal_downscaler and any other run type
2. All MEC/FFV2 rules in `verif_obs.smk` are wrapped in `if config["mec"] is not None:` so `config["mec"]["ekf_root"]` and `config["ffv2"][...]` are never accessed when those sections are absent
3. `ffv2_all` in `Snakefile` is wrapped in the same guard to prevent `rules.reorganize_ffv2_files` being referenced before it's defined

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -283,18 +283,20 @@ rule mec_all:
         ),
 
 
-rule ffv2_all:
-    """Target rule for FFV2 scoring pipeline (consumes fdbk_files, produces shiny scores)."""
-    input:
-        expand(
-            rules.reorganize_ffv2_files.output,
-            run_id=collect_all_candidates(),
-        ),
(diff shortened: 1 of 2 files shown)


clairemerker (approved):
I didn't try to run any workflow with the fix, I guess @jonasbhend did :) 
I think it looks fine, and we should probably merge fast to fix main. 

-> we need tests! and we can check when refactoring if this needs some further improvements :) 

Thanks for the fix!

PR #214, opened 2026-07-13
Allow selection of stations out of switzerland

lclanzi:
## Summary

This PR adds an optional `use_limitation` parameter to `fetch_data` in `jretrieve.py`. This allows to retrieve data from partner stations located out of Switzerland.

## Changes

- Add an optional `use_limitation` argument to `fetch_data`.
- Forward `--use-limitation <value>` to the `jretrievedwh.py` CLI when the parameter is provided, allowing station data to be filtered according to data usage rights.
- Preserve the current behaviour when `use_limitation` is omitted (`None`), allowing the CLI default value (`20`) to be used.

## Notes

- `use_limitation` is **not** added to `fetch_meta`, as the underlying meta-info endpoint does not support this option.
- additional information: https://service.meteoswiss.ch/jretrieve/api/v1/surface/help

### DIFF
--- a/src/data_input/jretrieve.py
+++ b/src/data_input/jretrieve.py
@@ -299,7 +299,7 @@ def fetch_data(
     increment_minutes=60,
     seq_type="surface",
     stage="prod",
-    use_limitation: int | None = None,
+    use_limitation: int = 40,
     timeout_s=600,
 ) -> pd.DataFrame:
     """Fetch observation data; columns: station (int), termin (YYYYMMDDhhmmss),
@@ -318,8 +318,7 @@ def fetch_data(
         "csv",
         *_stations_to_argv(stations),
     ]
-    if use_limitation is not None:
-        argv += ["--use-limitation", str(use_limitation)]
+    argv += ["--use-limitation", str(use_limitation)]
     LOG.info("jretrieve data: %s", " ".join(argv))
     return _parse_csv(_run_with_retry(argv, env=_build_env(stage), timeout_s=timeout_s))
 
--- a/src/data_input/jretrieve.py
+++ b/src/data_input/jretrieve.py
@@ -299,6 +299,7 @@ def fetch_data(
     increment_minutes=60,
     seq_type="surface",
     stage="prod",
+    use_limitation: int | None = None,
     timeout_s=600,
 ) -> pd.DataFrame:
     """Fetch observation data; columns: station (int), termin (YYYYMMDDhhmmss),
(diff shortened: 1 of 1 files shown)


jonasbhend:
Hi @lclanzi. Thanks for uncovering this. I think it would be ok to just set the use_limitation to 40 throughout. These are all the stations that we can use freely (as long as we don't publish the station data directly, which I don't see happening). This would cover all our use cases and be more in line with what users expect. When requesting SwissMetNet (i.e. 1,2) we still get the SwissMetNet stations only, if we include more measurement networks, we actually get the station data we request (which I understand is not the case now). If requesting by bounding box, you get all the stations in the area as expected. Thoughts?

lclanzi:
> Hi @lclanzi. Thanks for uncovering this. I think it would be ok to just set the use_limitation to 40 throughout. These are all the stations that we can use freely (as long as we don't publish the station data directly, which I don't see happening). This would cover all our use cases and be more in line with what users expect. When requesting SwissMetNet (i.e. 1,2) we still get the SwissMetNet stations only, if we include more measurement networks, we actually get the station data we request (which I understand is not the case now). If requesting by bounding box, you get all the stations in the area as expected. Thoughts?

I played a bit with the jretrieve API, and the only way I could retrieve stations outside Switzerland was by setting `use_limitation > 20`. Is your suggestion to change the dafault value of `use_limitation` to 40? That would make sense to me

jonasbhend:
> > Hi @lclanzi. Thanks for uncovering this. I think it would be ok to just set the use_limitation to 40 throughout. These are all the stations that we can use freely (as long as we don't publish the station data directly, which I don't see happening). This would cover all our use cases and be more in line with what users expect. When requesting SwissMetNet (i.e. 1,2) we still get the SwissMetNet stations only, if we include more measurement networks, we actually get the station data we request (which I understand is not the case now). If requesting by bounding box, you get all the stations in the area as expected. Thoughts?
> 
> I played a bit with the jretrieve API, and the only way I could retrieve stations outside Switzerland was by setting `use_limitation > 20`. Is your suggestion to change the dafault value of `use_limitation` to 40? That would make sense to me

Yes, exactly. Stations with usage limitation < 50 can be used for our activities, so this should be a meaningful default.

jonasbhend (approved):
Sorry for taking so long. This is an important addition, and given the current selection options (explicity list of stations, measurement networks, bounding box), it makes perfect sense to relax the restrictions on use_limitation from jretrieve.

PR #220, opened 2026-07-21
Fix retrieval of ICON constants

frazane:
This PR fixes the function used for fetching ICON constant GRIB files in the `src/data_input/__init__.py` which was assuming an incorrect schema in the response from the STAC API.


dnerini (approved):
Thanks for the fix!

clairemerker (approved):
Thanks!

PR #221, opened 2026-07-21
Add config-driven SAL precipitation verification

Louis-Frey:
Adds SAL (Structure–Amplitude–Location; Wernli et al. 2008) as a config-driven precipitation verification capability, mirroring how scorecards/scoremaps work: enabled via an `experiment.sal` block and computed for every requested run and baseline. Compute-only — plotting lives in the publication-figures workflow.

## What

- `src/verification/sal.py` — pysteps `sal()` wrapper + near-isotropic raster builder + nearest-neighbour remap (reuses `verification.spatial`). Dry windows (either field everywhere ≤ 0, NaNs counting as dry) return `(NaN, NaN, NaN)` instead of raising. Detection thresholds are the pysteps defaults (`thr_factor=0.067`, `thr_quantile=0.95`) as module constants — not config knobs.
- `workflow/scripts/verification_sal.py` — per-init compute driver (one job per participant/param/lead time). Scores the init times given by `--reftimes`, remaps both forecast and truth onto a common raster, and writes a **CSV** (one row per init: `reftime,S,A,L,fcst_mean,truth_mean` + a commented metadata header). Dry-window rows are retained with NaN S/A/L so a downstream wet-case filter can drop them. Every configured init must be present across forecast and truth — a missing one is a hard error, never a silent skip, so run and baseline are scored over an identical sample. SAL is a per-case scalar score, so a table is the natural container.
- `SalConfig` under `experiment.sal` — `enabled`, `params`, `leadtimes`, and the scoring raster (`grid_extent`, `grid_step_lat`, `grid_step_lon`). Validators reject non-precipitation params and malformed `grid_extent`; schema regenerated. The raster defaults to the ICON-CH1 analysis bounding box (`[-1.0, 18.0, 42.0, 50.5]`) at `0.01°` lat × `0.0145°` lon, i.e. ~1.1 km cells that are metrically square at 46.4°N (pysteps' Location term assumes square pixels).
- Merged the scoremaps and sal lead-time checks into a single `ConfigModel.validate_leadtimes_producible` model validator, which rejects lead times not produced by every participant for whichever of the two blocks is enabled.
- `verification_sal` / `verification_sal_baseline` Snakemake rules (output keyed by truth hash) + gated fan-out in `experiment_all`.
- Example `sal:` block added to all six example configs as `enabled: false`; everything else comes from the `SalConfig` defaults (`TOT_PREC6` at `[6, 12, 18, 24, 30]` h).
- pysteps + scikit-image dependencies (`pyproject.toml`, `uv.lock`).

## Example output

The CSVs are the only data product: one per (participant, param, lead time), at `data/runs/<run_id>/sal/<param>_<leadtime>_<truth_hash>.csv` (and `data/baselines/<baseline_id>/sal/…`). Below is `sal/TOT_PREC6_6_2b83.csv` from a five-init run — `TOT_PREC6` at `+6 h`, scored against the KENDA-CH1 analysis zarr:

```
# SAL (Wernli et al. 2008) per-init precipitation scores
# param: TOT_PREC6  accum_h: 6  step_h: 6  member: 000
# grid_extent: [-1.0, 18.0, 42.0, 50.5]  grid_cells: 851x1311
# source: .../output/data/runs/forecaster-b30a-4d02/07b7  n_init: 5
# reftime UTC YYYYMMDDHHMM; S/A/L are NaN for dry windows.
```

| reftime | S | A | L | fcst_mean | truth_mean |
|---|---|---|---|---|---|
| 2024-01-01 00Z | −0.103 | −0.221 | 0.079 | 0.745 | 0.930 |
| 2024-01-01 06Z | +0.378 | −0.627 | 0.186 | 0.153 | 0.293 |
| 2024-01-01 12Z | −0.145 | −0.717 | 0.258 | 0.051 | 0.107 |
| 2024-01-01 18Z | +0.574 | −0.225 | 0.027 | 0.629 | 0.788 |
| 2024-01-02 00Z | +0.221 | −0.137 | 0.018 | 1.219 | 1.398 |

(Values rounded for display; the CSV carries full float64 precision.)

Reading it: one row per initialisation, `S`/`A`/`L` dimensionless and signed, `fcst_mean`/`truth_mean` the domain-mean 6 h accumulation in mm over the scoring raster. Negative `A` throughout is a dry bias — this forecast under-predicts domain-total precipitation at every init here. The commented header pins everything needed to reproduce the numbers: the param and its accumulation period, the lead time, the baseline member, the raster extent and cell count, and the source archive.

## Testing

- Unit tests (`tests/unit/test_sal.py`): raster shape/bounds/orientation and custom extents, the remap gather and its NaN→0 fill, `compute_sal` on identical fields (all-zero triple) and on a dry window (all-NaN triple), `SalConfig` grid defaults matching the module constants, the grid and non-precip-param validators, the lead-time producibility validator, and the two script guards (lead time below the accumulation period, non-zarr truth).
- Integration longtest (`tests/integration/test_sal_small.py` + `configs/sal_small.yaml`): runs `evalml experiment` end to end on a minimal baseline-only config (ICON-CH2-CTRL, no inference — so no GPU, MLflow or DWH, only `/store_new` access) and checks the scores are *consistent*, not merely present — the documented metadata header, column set, raster size and one row per configured init; finite non-negative means; `A` equal to the identity recomputed from the row's own mean columns; S/A/L within their Wernli et al. bounds; S/L not NaN for every init; and — cross-checked across two independent output files — that the truth column depends only on valid time, which pins the `reftime + step` → accumulation-window mapping. All reference-free, so there is no blessed output to maintain. Marked `longtest`, so it is skipped by default and on GitHub Actions and runs on the CSCS balfrin runner.

## Notes

- Regridding is nearest-neighbour from the native (triangular ICON / KENDA) mesh onto the raster; SAL's S/A/L are unit-invariant, so only the reported means depend on the loader unit (mm).
- Keep `grid_extent` close to the truth's footprint. The remap has no distance cutoff, so cells beyond it merely repeat the nearest border value, and a larger raster also deflates pysteps' L term via its longer diagonal.
- Lead-time validation matches scoremaps: a short-horizon baseline (e.g. INCA `0/6/1`) restricts `sal.leadtimes` for all participants — use a config without such a baseline for longer lead times.
- Truth must be a gridded analysis zarr. Anything else — station observations (`jretrievedwh:…`) included — is rejected up front with a clear error, since remapping ~150 points onto the raster would yield meaningless scores.
- Scope: the longtest covers `verification_sal_baseline`. The run path (`verification_sal`) depends on `inference_execute`, so exercising it needs a GPU or the inference-replay fixture; the only untested delta is the `run_root/<reftime>/[grib]` resolution in `verification_sal.py`.


Louis-Frey:
Hi @clairemerker @frazane 

I created this PR to integrate core SAL functionality into evalML main. Would be great if one of you could have a look at this (probably not both needed, please decide yourself who is less swamped / more motivated ;-)). The plotting functionality for the paper that builds on this will follow later on in a PR on top of `paper-figures`. I already did a rather comprehensive review and test of this PR with Claude, but would for sure be good to have another human look at this as well.

Louis-Frey:
Hi @clairemerker @frazane, I boiled down this PR as much as I could / dared (because feedback last time was that the PR was too verbose). Would be great if one of you could have a look at it!

pyproject.toml
```diff
@@ -31,6 +31,8 @@ dependencies = [
     "earthkit-plots>=1.0,<1.1",
     "earthkit-meteo>=1.0,<1.1",
     "earthkit-geo>=1.0,<1.1",
+    "pysteps>=1.12",       # SAL spatial precipitation verification (salscores)
```
clairemerker:
```suggestion
    "pysteps>=1.12",
```

pyproject.toml
```diff
@@ -31,6 +31,8 @@ dependencies = [
...
+    "scikit-image>=0.22",  # required by pysteps.verification.salscores
```
clairemerker:
```suggestion
    "scikit-image>=0.22",
```

workflow/Snakefile
```diff
@@ -195,6 +196,28 @@ rule experiment_all:
             if SCOREMAPS_CONFIGS.get("enabled", False)
             else []
         ),
+        # SAL precipitation scores (compute-only: the CSVs themselves are the
+        # targets; plotting lives in the publication-figures workflow).
```
clairemerker:
```suggestion
        # SAL precipitation scores (compute-only: the CSVs themselves are the
        # targets.
```

src/evalml/config.py
```diff
@@ -265,6 +265,40 @@ class ScoreMapsConfig(BaseModel):
     model_config = {"extra": "forbid"}
 
 
+class SalConfig(BaseModel):
+    """SAL (Structure–Amplitude–Location; Wernli et al. 2008) per-init precip
+    scores, computed for every run and baseline when enabled and written to
+    per-participant CSVs. Compute-only; plotting lives in the figures workflow."""
```
clairemerker:
```suggestion
    per-participant CSVs."""
```

src/evalml/config.py
```diff
@@ -265,6 +265,40 @@ class ScoreMapsConfig(BaseModel):
...
+
+    enabled: bool = Field(
+        default=False,
+        description="Whether to compute SAL scores (requires pysteps + scikit-image).",
```
clairemerker:
```suggestion
        description="Whether to compute SAL scores.",
```
I think this info is not needed, except if theses packages are optional dependencies.

workflow/tools/config.schema.json
```diff
@@ -722,6 +734,46 @@
       "title": "Profile",
       "type": "object"
     },
+    "SalConfig": {
+      "additionalProperties": false,
+      "description": "SAL (Structure\u2013Amplitude\u2013Location; Wernli et al. 2008) per-init precip\nscores, computed for every run and baseline when enabled and written to\nper-participant CSVs. Compute-only; plotting lives in the figures workflow.",
```
clairemerker:
```suggestion
      "description": "SAL (Structure\u2013Amplitude\u2013Location; Wernli et al. 2008) per-init precip\nscores, computed for every run and baseline when enabled and written to\nper-participant CSVs.",
```

workflow/tools/config.schema.json
```diff
@@ -722,6 +734,46 @@
...
+      "properties": {
+        "enabled": {
+          "default": false,
+          "description": "Whether to compute SAL scores (requires pysteps + scikit-image).",
```
clairemerker:
```suggestion
          "description": "Whether to compute SAL scores.",
```

src/verification/sal.py
```diff
@@ -0,0 +1,87 @@
...
+# pysteps defaults (Wernli et al. 2008, eq. 1): detection threshold is
+# thr_factor * the thr_quantile-percentile of the wet precipitation.
+DEFAULT_THR_FACTOR = 0.067
+DEFAULT_THR_QUANTILE = 0.95
+
+# Below this truth point count SAL warns: a gridded analysis has millions of
+# points, a station network ~150, so the exact cut is not critical.
+MIN_TRUTH_POINTS = 10_000
+
+GRID_EXTENT = (-1.0, 18.0, 42.0, 50.5)  # lon_min, lon_max, lat_min, lat_max
```
clairemerker: Which domain does this correspond to?

workflow/scripts/verification_sal.py
```diff
@@ -0,0 +1,184 @@
...
+        else:
+            # A run stores GRIB under run_root/<reftime>/[grib/] (a missing init
+            # then fails in load_forecast_data).
+            src_root = args.run_root / reftime.strftime(DATETIME_FMT)
+            src_root = src_root / "grib" if (src_root / "grib").is_dir() else src_root
+
+        fcst = load_forecast_data(
+            src_root, reftime, [args.step], [args.param], member=args.member
+        )
+        if "step" in fcst.dims:
```
clairemerker: What happens in case there is no `step`?

workflow/scripts/verification_sal.py
```diff
@@ -0,0 +1,184 @@
...
+        truth_ds = load_truth_data(
+            args.truth, reftime, [args.step], [args.param], lazy_ds=truth_lazy
+        ).isel(time=0)
+
+        fcst_field, fcst_lat, fcst_lon = _native_1d(fcst[args.param])
+        truth_field, truth_lat, truth_lon = _native_1d(truth_ds[args.param])
+
+        if fcst_idx is None:
+            # SAL needs a resolved (gridded) truth field; warn if it looks sparse.
+            if truth_lat.size < MIN_TRUTH_POINTS:
```
clairemerker: Do we need to check for that? `open_truth_zarr()` calls `_open_analysis_zarr()`, I would expect it not to be able to load observational data?

workflow/scripts/verification_sal.py
```diff
@@ -0,0 +1,184 @@
...
+            s,
+            a,
+            ell,
+            fcst_mean,
+            truth_mean,
+        )
+        rows.append((reftime.strftime(DATETIME_FMT), s, a, ell, fcst_mean, truth_mean))
+
+    # One row per init; fixed metadata in a commented header (pandas skips it via
+    # read_csv(comment="#")). S/A/L are NaN for dry windows.
```
clairemerker:
```suggestion
    # One row per init; fixed metadata in a commented header.
    # S/A/L are NaN for dry windows.
```

clairemerker (commented):
Thanks, I think it looks fine :) 
I added some comments to the changes, and I wonder if you should already include an integration test config for it?

Louis-Frey:
Thanks for the feedback @clairemerker, I will look into it. 

Yes, good idea, I will include a longtest as well.

workflow/scripts/verification_sal.py
Louis-Frey: Not sure if I get the point. `truth_lat` (which is checked) is derived from `truth_ds`, which comes out of `load_truth_data`, which has both a .zarr branch and a jretrieve branch. So at least at the script level, `truth_lat` / `truth_ds` could contain observational data. Maybe this never occurs during the normal evalml workflow, but I feel like the check is quite light and therefore justifiable as defense-in-depth.

clairemerker: Ah I see, I thought the code checked for the root to be a zarr when trying to use lazy loading with `truth_lazy`, but it doesn't, it just sets it to `None` in case it's not a zarr. I personally feel like it would be nicer to make sure no station data feeds in ealier than here (e.g. by catching "jretrieve" in `root` early in the script), also because the min number of points heavily depends on the resolution and the domain (shouldn't we also check for the extend of the covered domain, otherwise the remap could look weird)?

workflow/scripts/verification_sal.py
Louis-Frey: As far as I can see, there is no such case — but good prompt to actually check, 
because the conditional was dead code and I've now removed it.

`step` is always a dimension of the loaded forecast, on every path:

- ML run GRIB and the ICON archive both go through `_load_forecast_data_from_grib` →
  `load_from_grib_file` → `to_xarray(**XARRAY_ENGINE_PROFILE)`, and that profile sets
  `ensure_dims: ["z", "number", "step", "forecast_reference_time"]`. earthkit
  guarantees such a dimension is preserved even when `squeeze=True` and its size is 1
  — which is precisely our case, since we request a single step.
- The INCA NetCDF loader ends with `swap_dims({"valid_time": "step"})`.
- Nothing downstream squeezes it: `_disaggregate_accum` and
  `_disaggregated_and_derived_params` only use list-form `.sel(step=[...])`, which
  keeps the dimension.

So the guard could never be False, and its `else` path silently passed the field
through without verifying the lead time. Replaced with an unconditional
`fcst = fcst.sel(step=np.timedelta64(args.step, "h"))`. If the invariant is ever
broken upstream this now fails loudly and immediately — `KeyError` if `step` is
absent, `ValueError` if it degenerates to a scalar coordinate — instead of scoring an
unverified field and labelling it with the requested lead time in the CSV header.

Note `verification_scoremaps.py` carries the same pattern (predates this branch);
leaving it alone here to keep the diff scoped, happy to follow up separately.

Louis-Frey: Addressed in [2c63609](https://github.com/MeteoSwiss/evalml/pull/221/commits/2c63609fc174a232688e2c6c13b9ba7459bf081a)

workflow/scripts/verification_sal.py
Louis-Frey: Yes, makes sense to check earlier and fail fast. Changed it in commit [6e93e57](https://github.com/MeteoSwiss/evalml/pull/221/commits/6e93e57b88d2b4a482f9cbe6be946ecb8f61d196). I temporarily had a solution checking the point density, but removed it again in the interest of brevity. 

About the domain, this is a fair point. I also had a more flexible solution that allowed for a user-defined target grid, but abandoned this as well in the interest of brevity in favour of the current hard-coded grid. It's a single commit that I could most likely easily revert, should I do it?

src/verification/sal.py
Louis-Frey: Right now it doesn't correspond to any pre-defined domain, which I admit is sub-optimal. Also regarding the other comment, I could make the target grid construction based on one of the predefined domains, and give an option for a completely user-defined grid. What do you think?

Louis-Frey:
Added a longtest in [74b3be2](https://github.com/MeteoSwiss/evalml/pull/221/commits/74b3be280e815165ac44ee3965802f56d2e313a3).

src/verification/sal.py
Louis-Frey: Made the domain and resolution configurable again, in [9359b26](https://github.com/MeteoSwiss/evalml/pull/221/commits/9359b26c7540d8af1974997712cafd659093d9a9).

PR #222, opened 2026-07-21
Fix meteogram plotting

clairemerker:
AI assisted implementation!

- Correct the plot_meteogram import from the nonexistent data_import module to data_input
- Drop the unused DWH station-catalog fetc
- Run the plot rule under 'uv run python' so it uses the project environment always
- Add a longtest that drives 'evalml showcase' on a minimal config and asserts a meteogram PNG is produced

I ran the new longtest locally, it passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,9 @@ dev = [
 ]
 
 [tool.pytest.ini_options]
+# Skip longtests (GPU/credential/data-dependent integration tests) by default,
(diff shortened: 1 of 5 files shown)


clairemerker:
The CI is also running longtests... I missed that, will fix it later :)

clairemerker:
CI is fixed :) I think we can discuss how to handle long tests and when to run them in the next days!

frazane (approved):
Thank you!

PR #231, opened 2026-07-24
Add integration test for configs

lclanzi:
Adds an integration test that runs the full evalml experiment pipeline and asserts that key verification metrics match reference values from a known-good run.

### DIFF
--- a/ci/cscs.yml
+++ b/ci/cscs.yml
@@ -40,7 +40,7 @@ integration_test_job:
       UV_HOME="$VENV_DIR" uv sync
       source "$VENV_DIR/bin/activate"
   script:
-    - pytest tests/integration -m longtest
+    - pytest tests/integration -m heavytest
   variables:
     # Explicitly request more time from Slurm, so that the long tests can complete.
     SLURM_TIMELIMIT: "10:00:00"
--- a/ci/cscs.yml
+++ b/ci/cscs.yml
@@ -43,4 +43,4 @@ integration_test_job:
     - pytest tests/integration -m longtest
   variables:
     # Explicitly request more time from Slurm, so that the long tests can complete.
-    SLURM_TIMELIMIT: "01:00:00"
+    SLURM_TIMELIMIT: "10:00:00"
--- a/resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper-test.yaml
+++ b/resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper-test.yaml
@@ -1,0 +1,81 @@
+lead_time: 12h
+write_initial_state: true
+allow_nans: true
+
+env:
+  ANEMOI_INFERENCE_NUM_CHUNKS: 8 # OOM error if not set
+
+# inputs
+input:
+  test:
+    use_original_paths: true
+
+post_processors:
+  - accumulate_from_start_of_forecast: # accumulate tp from start of forecast
+      accumulations:
+        - tp
+  - forward_transform_filter:
+      rescale:
+        scale: 1000 # convert units from m to kg m-2
+        offset: 0
+        param: tp
+
+output:
+  tee:
+    - grib:
+        path: grib/{date}{time:04}_{step:03}.grib
+        encoding:
+          typeOfGeneratingProcess: 2
+          centre: lssw
+        templates:
+          samples: resources/templates_index_icon.yaml
+        post_processors:
+          - extract_mask: # removes global points
+              mask: "lam_0/cutout_mask"
+              as_slice: true
+    - grib:
+        path: grib/ifs-{date}{time:04}_{step:03}.grib
+        encoding:
+          typeOfGeneratingProcess: 2
+          centre: ecmf
+        templates:
+          samples: resources/templates_index_ifs.yaml
+        post_processors:
+        - extract_mask: # removes lam points
+            mask: "lam_0/cutout_mask"
+            as_slice: true
+            inverse: true
+        - assign_mask: # fill local/global overlapping points with nan
+            mask: "global/cutout_mask"
+        modifiers:
+          - patches:
+            - variable:
+                U_10M: {"param": 165, "shortName": "10u"}
+                10u: {"param": 165, "shortName": "10u"}
+                V_10M: {"param": 166, "shortName": "10v"}
+                10v: {"param": 166, "shortName": "10v"}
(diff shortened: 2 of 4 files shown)


ci/cscs.yml
jonasbhend, line 1: I suggest to add the heavy tests in a separate ci config file for which we can set up a cron-triggered pipeline (or trigger with an alternative command). What do you think @clairemerker ?

```
ci/cscs.yml        # current configuration with `-m longtest`
ci/cscs_cron.yml   # copy of pipeline from cscs.yml with `-m heavytest`
```

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper-test.yaml
jonasbhend, line 1: Do we need that? Can't we reuse the existing inference config?

tests/integration/configs/varda-single-1.0.yaml
```diff
@@ -0,0 +1,131 @@
...
+      root: /store_new/mch/msclim/INCA
+      steps: 0/6/1
+  - baseline:
+      label: ICON-CH2-CTRL
+      root: /store_new/mch/msopr/osm/ICON-CH2-EPS
+      steps: 0/12/1
+  - baseline:
+      label: ICON-CH1-CTRL
+      root: /store_new/mch/msopr/osm/ICON-CH1-EPS
+      steps: 0/12/1
```
jonasbhend, line 34: maybe make all baselines different length (as in reality but shorter)?
```suggestion
      steps: 0/9/1
```

tests/integration/configs/varda-single-1.0.yaml
```diff
@@ -0,0 +1,131 @@
...
+    - TOT_PREC1
+    - TOT_PREC6
+    - PMSL
+  stratification:
+    regions:
+      - mittelland
+      - berge
+      - alpennordseite
+      - alpensuedseite
+      - jura
```
jonasbhend, line 59: here we could probably restrict to two regions only

tests/integration/configs/varda-single-1.0.yaml
```diff
@@ -0,0 +1,131 @@
...
+  thresholds:
+    TOT_PREC1:
+      gt: [0.0, 0.1, 1, 5, 10]
+    TOT_PREC6:
+      gt: [0.0, 0.1, 5, 10, 20, 50]
+    SP_10M:
+      gt: [2.5, 5.0, 10.0]
+    T_2M:
+      lt: [273.15]
+      gt: [288.15, 298.15]
```
jonasbhend, line 70: Also here, we could limit the number of thresholds to 2 per case (for precip 0 and something large)

tests/integration/test_configs.py
```diff
@@ -0,0 +1,81 @@
...
+        "PMSL.MAE":       pytest.approx(32.933578, abs=1e-6),
+        "PMSL.CORR":      pytest.approx(0.928961, abs=1e-6),
+        "TOT_PREC6.RMSE": pytest.approx(0.450449, abs=1e-6),
+        "TOT_PREC6.BIAS": pytest.approx(0.077970, abs=1e-6),
+        "TOT_PREC6.MAE":  pytest.approx(0.112071, abs=1e-6),
+        "TOT_PREC6.CORR": pytest.approx(0.720550, abs=1e-6),
+    },
+}
+
+SEL = {"region": "all", "season": "all", "init_hour": -999}
```
jonasbhend, line 48: Maybe we need to select something non-standard here?

jonasbhend (commented):
shoot, I had a bunch of comments that were left hanging :-)

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper-test.yaml
lclanzi: When you do testing, you may have to modify the inference configs. What about moving this config into the test folder?

tests/integration/test_configs.py
lclanzi: I took that config just for testing this idea. I think now we can select ad-hoc configs for ideally having a 100% coverage

resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper-test.yaml
jonasbhend: I was hoping that we could also test the inference configs that we use for real-live evaluation. So I would strongly encourage to not do this by default and I still don't see in what cases we would need to change the inference configs for testing.

jonasbhend: and yes, of course if we really need to modify the configs, lets move them to tests/... but to me this defeats a bit the purpose of the integration tests. Having said that, the same applies to the 'example' configs. Ideally we could reuse the ones from the main config folder instead of creating new ones for testing. That way we could also guarantee that all use cases are represented in example configs (because we can test for coverage), which is not the case today.

lclanzi: In this specific case, I only had to change the lead time from 120h to 12h. But yes, what you say makes sense.

jonasbhend: I guess that change would also have made sense for the example config. I use these mostly for development / testing where indeed we also profit from shorter feedback loops.

jonasbhend: So really the question is if we shouldn't use the configs in ./config for the heavytest as initially discussed. That way we have a real integration test and also the benefit of being able to profit from complete coverage.

clairemerker: Hm... I think for the specific "shorter" tests that are not really representing a real-life case having them in the test folder is fine. 
For the others, could it be an idea for the tests to copy them from the config folder and on the fly override the important blocks (time period, variables, domains,...)?

clairemerker: (ideally the settings in the config folder would be ok for the integration tests, but if developers are not aware or forget about it we might have bad surprises on the weekly run :D )

PR #232, opened 2026-07-24
Small integration test for meteogram

jonasbhend:
This PR implements minor changes to the existing meteogram integration test.

### Summary of changes
* rename test and config to meteogram_small
* change date to 2024-08-01 (official start of ICON archive)
* additional parameters to cover all use cases (derived, aggregated and instantaneous)


tests/integration/test_meteogram_small.py
```diff
@@ -34,13 +35,17 @@ def test_showcase_meteogram():
...
+        for station in EXPECTED_STATIONS:
+            pngs = glob.glob(
+                str(
+                    PROJECT_ROOT
+                    / f"output/results/**/202408010000_{param}_{station}.png"
+                ),
+                recursive=True,
+            )
+            assert pngs, (
+                f"expected meteogram PNG (202408010000_{param}_{station}.png) was not produced"
```
dnerini: I would argue that there is no need to check for output files, since this is done by snakemake and the run will fail if declared outputs are missing

jonasbhend: Good point. That makes the whole testing quite a bit easier.

jonasbhend: done

clairemerker (approved):
Nice! thanks!

marymcglo:
cscs-ci run

marymcglo:
Ignore the new test failure: we switched to service account, and it is not working yet. (Test was passing before)

PR #234, opened 2026-07-24
Use /dev/shm for environment creation and always squash venv

jonasbhend:
Squashing the venv from the filesystem takes very long when the filesystem is under heavy load. To speed up the process and reduce load on the filesystem, this PR implements RAM-based creation of the squashfs image. Thanks @huppd for the suggestion.

### Summary of changes
* use `/dev/shm` instead of the `OUT_ROOT` for creating the venv
* clean up `/dev/shm` after creation
* remove config option to avoid squashing for simplicity

### DIFF
--- a/src/evalml/config.py
+++ b/src/evalml/config.py
@@ -101,12 +101,6 @@ class RunConfig(BaseModel):
         False,
         description="If true, the ECCODES_DEFINITION_PATH environment variable will not be set to the COSMO local definitions.",
     )
-    squash_venv: bool = Field(
-        True,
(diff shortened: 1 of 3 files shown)


jonasbhend:
According to my tests, venv creation took:

Phase | Duration
-- | --
venv create + pip install | 29 s
bytecode compilation | 13 s
mksquashfs | 9 s
Total | 51 s

clairemerker (approved):
Didn't get to test it yet but the code looks good! Thanks a lot!

lclanzi (approved):
Thanks!

PR #235, opened 2026-07-24
add environment installation to CI/CD

marymcglo:
Under the service account, uv needs to be installed for every CI run since there is no global installation.


marymcglo:
cscs-ci run

PR #236, opened 2026-07-26
Fix for issues discovered during evaluation

jonasbhend:
This PR brings a few fixes to main that have been discovered during recent evaluation runs.

### Summary of changes
* enable lapse-rate correction for AIFS
* use post-processor that is consistent with anemoi-inference version for rescaling of precipitation with AIFS
* add missing predefined regions for keyword-based definition

### DIFF
--- a/resources/inference/configs/aifs-single-forecaster.yaml
+++ b/resources/inference/configs/aifs-single-forecaster.yaml
@@ -14,7 +14,7 @@ post_processors:
   - accumulate_from_start_of_forecast:
       accumulations:
         - tp
- # We use `backward_transform_filter` instead of `forward_transform_filter` 
+ # We use `backward_transform_filter` instead of `forward_transform_filter`
  # as this is not supported in the pinned anemoi-inference version used for aifs-single's evalml config (anemoi-inference=0.11.1)
   - backward_transform_filter:
       filter: rescale
--- a/resources/inference/configs/aifs-single-forecaster.yaml
+++ b/resources/inference/configs/aifs-single-forecaster.yaml
@@ -14,6 +14,8 @@ post_processors:
   - accumulate_from_start_of_forecast:
       accumulations:
         - tp
+ # We use `backward_transform_filter` instead of `forward_transform_filter` 
+ # as this is not supported in the pinned anemoi-inference version used for aifs-single's evalml config (anemoi-inference=0.11.1)
   - backward_transform_filter:
       filter: rescale
       scale: 0.001 # convert units from m to kg m-2 (backward: x/scale = x*1000)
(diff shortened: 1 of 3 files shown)


resources/inference/configs/aifs-single-forecaster.yaml
```diff
@@ -14,11 +14,11 @@ post_processors:
...
-  - forward_transform_filter:
-      rescale:
-        scale: 1000 # convert units from m to kg m-2
-        offset: 0
-        param: tp
+  - backward_transform_filter:
+      filter: rescale
+      scale: 0.001 # convert units from m to kg m-2 (backward: x/scale = x*1000)
+      offset: 0
+      param: tp
```
MicheleCattaneo, line 24: I suggest to add a comment to explain why we use a backward transform, which is less intuitive than a forward transform
```suggestion
 # We use `backward_transform_filter` instead of `forward_transform_filter` 
 # as this is not supported in the pinned anemoi-inference version used for aifs-single's evalml config (anemoi-inference=0.11.1)
  - backward_transform_filter:
      filter: rescale
      scale: 0.001 # convert units from m to kg m-2 (backward: x/scale = x*1000)
      offset: 0
      param: tp

```

workflow/rules/common.smk
```diff
@@ -86,12 +86,21 @@ def parse_regions():
...
         if isinstance(entry, str):
-            result.append({"type": "shp", "name": entry, "path": f"{root}/{entry}.shp"})
+            if entry in PREDEFINED_REGIONS:
+                result.append(
+                    {"type": "bbox", "name": entry, "bbox": PREDEFINED_REGIONS[entry]}
+                )
+            else:
+                result.append(
+                    {"type": "shp", "name": entry, "path": f"{root}/{entry}.shp"}
+                )
```
MicheleCattaneo, line 111:
```suggestion
            if entry in PREDEFINED_REGIONS:
                logging.getLogger("snakemake").info("Region %s found among the predefined regions (rectangles in lon/lat).  The evaluation will happen on the bounding box %s", entry, PREDEFINED_REGIONS[entry])
                result.append(
                    {"type": "bbox", "name": entry, "bbox": PREDEFINED_REGIONS[entry]}
                )
            else:
                path = f"{root}/{entry}.shp"
                logging.getLogger("snakemake").info("Region %s not found in the predefined regions. Will try to locate file %s.", entry, path)
                result.append(
                    {"type": "shp", "name": entry, "path": path}
                )
```

MicheleCattaneo (commented):
I suggest to explicitly log where the region is taken from or tried to be taken from. Easier to debug in the future.

PR #237, opened 2026-07-27
split CI into two pipelines

clairemerker:
Two different pipelines:
- longtest: intergration tests
- heavytest: heavy integration tests (full end to end)

The idea is to trigger heavytest weekly or nighly using the CI cron, and longtest in PRs using the comment call.

### DIFF
--- a/ci/heavytest.yml
+++ b/ci/heavytest.yml
@@ -1,0 +1,14 @@
+include:
+  - ci/cscs.yml
+
+longtest:
+  extends: [.baremetal-runner-balfrin]
+  before_script:
+    - !reference [.sh_preamble, script]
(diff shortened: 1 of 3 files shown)


clairemerker:
cscs-ci run longtest

ci/heavytest.yml
```diff
@@ -0,0 +1,14 @@
+include:
+  - ci/cscs.yml
+
+longtest:
```
Louis-Frey: Should this be called "longtest" or rather "heavytest"?

clairemerker: Oh, copy-paste mistake :D Thanks!

PR #239, opened 2026-07-27
Feat/test inference fixture

clairemerker:
AI assisted implementation! 

Adds an opt-in mechanism to replay pre-computed inference output instead of running inference, so the integration longtest (currently test_showcase_meteogram) runs wihtout needing inference. 
Truth and baselines still come from the DWH and /store_new: only inference is frozen.

Adds a new top-level config key fixture_root (absent by default → production unchanged):

- unset: inference_execute runs anemoi-inference exactly as previously
- set: inference_execute symlinks frozen GRIB from <fixture_root>/data/runs/<run_id>/<init_time>/grib into the run workdir and marks the step done

The fixture mirrors the pipeline's own output/data/runs/... layout, so capture and replay agree by construction (guarded by a round-trip unit test)

One-time capture (GPU node, needs MLflow/DWH//store_new):
```
# comment out fixture_root in tests/integration/configs/meteogram_small.yaml
evalml showcase tests/integration/configs/meteogram_small.yaml
evalml capture-fixture tests/integration/configs/meteogram_small.yaml \
    /store_new/mch/msopr/cmerker/evalml_test_fixtures/meteogram-small
# restore fixture_root
```

Replay (fast, no GPU) — the default once the fixture exists:

```
pytest tests/integration -m longtest
```

If `fixture_root` is set but the fixture is missing, the run fails fast with a clear message pointing at evalml capture-fixture.


Louis-Frey:
AI assisted review! ;-)  For a first iteration it should do. 
Find Claude's assessment below. I told Claude to be rigorous and critical in the assessment, but polite in the feedback, hope it worked ^^

---------------------------------------------
Nice feature — freezing only inference while keeping truth and baselines live is the right boundary. I reviewed head `aa3545b` and also exercised the mechanism locally, since it's hard to judge from the diff alone; findings marked **(verified)** were reproduced rather than inferred. Three things I'd like fixed before merge, plus a few smaller ones.

### Checked and working

- **The fixture path is invariant to `fixture_root`** — `run_id` derives only from the env/run hashes, so the "comment out, capture, restore" sequence really does produce matching paths. Recomputing the hashes from the committed `meteogram_small.yaml` gives `forecaster-b30a-4d02/864a`, exactly the directory in the fixture. **(verified)**
- **Leaving fixture mode doesn't corrupt the fixture.** Snakemake's `remove_existing_output()` unlinks the symlink rather than following it, so anemoi never writes through it into `/store_new`. **(verified)**
- `snakefmt --check` passes, `config.schema.json` matches a fresh regeneration, all 10 new unit tests pass.

### Please fix before merge

**1. `capture-fixture` deletes the fixture when run after a replay (verified).** `src/evalml/fixtures.py:40-45` — after a replay, `output/data/runs/<run_id>/<init>/grib` is a symlink into the fixture, and `iter_grib_dirs` follows it, so `dest` resolves to the symlink's own target: `shutil.rmtree(dest)` removes the fixture, then `copytree` fails on a now-dangling source. I reproduced this end to end (`FileNotFoundError`, fixture gone). The README does say to comment out `fixture_root` first, but 8.5 GB plus a GPU node is a steep price for an ordering slip. Quieter variant: capturing to a *new* `fixture_root` while the symlink is in place copies the *old* fixture forward with a fresh `captured_at`, so a refresh can appear to succeed without refreshing anything. Both go away if `iter_grib_dirs` skips symlinks, or if `capture_fixture` returns early when `grib.resolve() == dest.resolve()`.

**2. Capture isn't scoped to the config (verified).** It globs everything under `output/data/runs`, so the fixture on `/store_new` contains five GRIB dirs for a config with one forecaster and one date — 8.5 GB total, of which ~7.6 GB is a `forecaster-sruc` / `temporal_downscaler` experiment that shared the same `output/`. The manifest's single `checkpoints` entry then implies all five came from `b30a`, which makes the provenance record misleading. Filtering capture to the run IDs and dates of the given config fixes both.

**3. The replay rule `rm -rf`s a directory Snakemake owns.** `workflow/rules/inference.smk:292` — that path is a declared `directory()` output of `inference_prepare_forecaster`, but fixture-mode `inference_execute` declares only the `.ok` file. The guard is needed (`ln -sfn` otherwise nests the link inside the existing directory), but in the README's own sequence the first replay silently deletes the real inference output, and as a shell side effect it never shows up in `--dry-run`. `rmdir` (fails on non-empty) or an explicit "move it aside" error keeps the benefit without the risk.

### Also worth addressing

**4. Test docstring and inference coverage.** `tests/integration/test_meteogram_small.py:14-20` still says the test drives inference end to end and needs a GPU and MLflow, which now contradicts the new `tests/integration/README.md` — please update it. More importantly, this was the only CI job running `anemoi-inference`, and nothing does now; `earthkit-utils` / `earthkit-data` remain unpinned ranges, so that drift would go unnoticed. Could a real-inference run be kept as a weekly `heavytest` (`ci/heavytest.yml` already exists)?

**5. Fixture location and refresh burden.** `run_specific_hash` includes the contents of `resources/inference/configs/sgm-forecaster-global-ich1.yaml`, a shared file that has changed five times. Any PR touching it changes `run_id` and the longtest fails with `Fixture GRIB not found` — recoverable only by someone with a GPU node *and* write access to `/store_new/mch/msopr/cmerker/`. Please move the fixture somewhere team- or CI-owned and reference it from the config by a non-personal path. Related: the `grib` dirs are `drwxr-x---+` (group `s83`), and since `exists()` only needs traverse on the parents, the fail-fast check passes for a non-member while the read fails later inside plotting.

**6. Replay should be visible in the logs.** `VERIF_HASH` doesn't include `fixture_root`, so a replayed `verif_<hash>.nc` has the same path as a real one. A line in the `onstart` banner (`Inference: REPLAYED FROM FIXTURE <path>`) would make it unmistakable. Consider also passing it via `--config fixture_root=…` from the pytest fixture instead of committing it to the YAML — that removes the comment-out/restore dance and keeps a test-only switch out of the production config model.

**7. Test the branch that carries the risk.** `tests/unit/test_fixtures.py:47-58` checks `capture_fixture` against `fixture_grib_dir` — two functions in the same module, over a layout the test builds itself; the layout that must stay in sync lives in `workflow/rules/inference.smk:181-183`. A `snakemake -n` test with a small config and a fake fixture (replay rule selected, env-build rules absent) plus a negative test for the missing-fixture `WorkflowError` would cover the feature itself.

**8. A relative `fixture_root` yields a dangling symlink (verified)** — `exists()` resolves against the launch cwd, `ln -sfn` against `workdir`. A `.resolve()` in `fixture_grib_dir` rules it out.

### Minor

- `MANIFEST.yaml` isn't read by anything; GRIB checksums plus the evalml commit, validated at replay, would catch a checkpoint replaced behind an unchanged MLflow URI.
- `verif_obs.smk:88` takes the GRIB dirs as real inputs, so with `mec` enabled fixture mode still triggers the checkpoint download — worth documenting as out of scope.
- The `srun`/`squashfs-mount` block now exists in two copies in the `if`/`else`; a conditional `shell` (or `ruleorder`) would keep the production path single-sourced.
- `copytree` preserves the source mtime, so the fixture is older than the `.ok` file from the capture run; an `os.utime` on capture makes rerun behaviour deterministic.
- `cli.py:263` bypasses `load_yaml` and `ConfigModel` (CLI and workflow could disagree on `output_root`) and re-imports `yaml` locally; `cli.py:272-276` misses the nested `forecaster.checkpoint` of a `temporal_downscaler` and uses a naive `datetime.now()`.
- `inference.smk:286`: `fixture_grib=lambda wc, input: input.grib` is redundant — `{input.grib}` works directly.
- `fixture_grib_dir` is imported in `common.smk:12` but only used in `inference.smk`.
- CI hasn't run here (0 checks, status `pending`): `.github/workflows/ci.yaml` only triggers on PRs to `main`.
- PR description: "runs needing inference" → "runs *without* needing inference"; `meteogram_small.yaml` still points at the renamed `test_showcase.py`.

clairemerker:
Thanks for the review! 
I addressed all of it.

clairemerker:
cscs-ci run longtest

clairemerker:
cscs-ci run longtest

clairemerker:
cscs-ci run longtest

clairemerker:
cscs-ci run longtest

clairemerker:
cscs-ci run longtest

PR #243, opened 2026-07-28
Heavytest forecaster

Louis-Frey:
# Enable forecasters-ich1 in the heavytest integration suite

Adds `forecasters-ich1.yaml` to the heavytest metric check together with its generated expected-metrics file, extends the check to configs that produce more than one forecaster run, replaces the commented-out reference generator with a standalone script that shares its plumbing with the test, covers that script with unit tests, and fixes a CI-only failure in the plotting stage that made this config unrunnable.

## Changes

| File | Diff | Notes |
| --- | --- | --- |
| `tests/integration/expected/forecasters-ich1.yaml` | +7466 | Generated reference — 2 forecaster runs × 36 sel rows × 90–102 metrics = 7104 values |
| `tests/unit/test_regenerate_expected.py` | +271 | New — 18 tests over the generator and both comparison formats |
| `tests/integration/regenerate_expected.py` | +174 | New — the generator, as a script |
| `tests/integration/configs/forecasters-ich1.yaml` | +122 / −21 | The config being enabled |
| `tests/integration/test_configs.py` | +84 / −103 | Multi-source comparison; plumbing and generator moved out |
| `tests/integration/expected_metrics.py` | +65 | New — plumbing shared by test and generator; `CONFIGS` lives here |
| `workflow/scripts/prefetch_natural_earth.py` | +72 | New — Natural Earth cache warm-up |
| `tests/integration/README.md` | +48 | How to regenerate a reference |
| `workflow/rules/plot.smk` | +17 | Wires the warm-up ahead of the plot fan-out |

The generated reference is 90% of the diff; the reviewable surface is ~850 added and 124 removed lines.

## The config

Eight init times across April and June: two seasons, two init hours, and two inits in each bucket, so aggregation is exercised rather than just executed. Two forecasters against an ICON-CH2-EPS baseline, 8 parameters, 4 regions, thresholds per parameter, scorecards on with three sections (region, season, init_hour), scoremaps on. One run is 1725 snakemake steps, 1620 of them scoremap plots.

## Multi-source metric checking

A config with several forecasters writes several `verif_aggregated_*.nc` files, and the previous check only looked at `nc_files[0]` and its first non-truth source — so additional runs were silently unverified. Expected entries are now keyed by source hash-prefix, here `forecaster-b30a-4d02` and `forecaster-973b-7e40` for the two checkpoints, and every non-truth source found across all output files is compared.

`assert checked_sources == set(expected)` guards against a vacuous pass: if a run's outputs are missing a source the reference expects, the test fails rather than quietly checking less. Mismatches are collected rather than raised on first hit, so a failure reports full pass/fail counts, diff range and median, and per-variable failure counts instead of a single value — extending the single-source reporting from base #245 to the keyed layout.

The tolerance stays at base #245's `1e-1`, absolute; only the comment is new, recording the measured drift that justifies it (see test results).

The `isinstance(expected, list)` branch is deliberate back-compat for `varda-single-1.0.yaml`, whose reference is still in the older flat-list format. It can be deleted once that file is regenerated, which needs operational DWH access for varda's truth.

## Regenerating a reference

The generator used to be a commented-out block run as `python tests/integration/test_configs.py`. Being dead code it had drifted: it looped over `CONFIGS` but globbed all of `output/data/runs/**` inside the loop, so every config would have been written the same content, and it called `math.isfinite()` without importing `math`.

It is now a script, taking exact config names:

```
python tests/integration/regenerate_expected.py forecasters-ich1.yaml
```

Name several configs to regenerate several, or pass `all`. `test_configs.py` only asserts, so CI cannot rewrite a reference by any combination of flags.

An earlier revision of this PR made it a `--regenerate-expected` pytest flag; review feedback moved it out into the script, which deleted 108 lines of `tests/conftest.py` rebuilding guard rails argparse gives for free — a required positional replaces "a bare flag is rejected", `choices=` replaces hand-rolled exact-name matching, and the "you forgot `-m heavytest` and it went green having regenerated nothing" guard turned out to be a failure mode that only existed because the generator was a test.

It is not a copy-paste of the comparison logic, which is what let the old block rot. `expected_metrics.py` holds what both paths need — `find_nc_files()`, `run_experiment()`, `run_sources()`, `source_key()`, `metric_value()` — and both import it, so a reference cannot be generated with a different selection, or from a different experiment invocation, than the one it is asserted against. That extraction also removed three inlined copies of the truth-source filter and source-key derivation from the comparison path.

Two sharp edges remain, both aborting before anything is written. Exact names only: `varda-single` is rejected with the valid names listed, because it would also match a future `varda-single-2.0.yaml` at hours of GPU per config. And only the `.nc` files the run rewrote are used, tracked by snapshotting mtimes before the experiment — `output/data/runs/` is shared across configs, so a plain glob would write one config's runs into another's reference. If snakemake considered the outputs up to date and rewrote nothing, that is reported with the fix rather than silently regenerating from a stale tree. The glob is sorted so a multi-run config's keys come out in the same order on any machine.

## Unit tests for the generator

Moving the generator out of `test_configs.py` also moved it out of everything that ran it, so `tests/unit/test_regenerate_expected.py` drives it against a synthetic `verif_aggregated`-shaped dataset: source keying and the exclusion of truth sources, stat suffixes and non-finite scores; coverage of every source across several files; all four abort paths; the argparse guards; and, on the comparison side, both reference formats, drift beyond tolerance, and the missing-source guard. Written values are checked against an independently computed mean over lead time rather than through `metric_value()`, which would be a tautology since the generator reads through it too.

18 tests, in the default suite (143 collected), ~48 s locally. Mutation-checked rather than merely green — that exposed a real hole, fixed in the same commit: `setdefault()` creates a source's key before any row is built, so an all-NaN run produced `{source: []}`, which is truthy, and the "refusing to write an empty reference" guard could not fire.

## Prerequisite: Natural Earth cache warm-up

Enabling this config is what first ran 1620 scoremap plot jobs in parallel, which exposed a pre-existing race. The plots draw coastlines via earthkit-plots → cartopy, which downloads Natural Earth shapefiles into a cache under `$HOME` on first use. `ci/cscs.yml` sets `HOME="$PWD/home"`, recreated fresh every pipeline, so that cache is always cold in CI: the plot jobs all start at once, race to download the same files, and one reads a half-written shapefile and dies with `struct.error: unpack requires a buffer of N bytes`, aborting the workflow at ~5% completion.

The new `plot_prefetch_natural_earth` localrule downloads and fully reads the three 50m datasets the plots use, once and serially, retrying up to three times, and both scoremap rules take its marker as an input. Local runs reuse a persistent `$HOME` and so were never affected, which is why this only ever surfaced in CI.

## Test results

`forecasters-ich1.yaml` **passes**. Two full heavytest CI runs on this branch: 2026-08-03 (at `4195a24`, comparison mode) green against the committed reference, and 2026-08-06 completing all 1725 steps.

**Measured run-to-run drift.** Regenerating the reference from the 2026-08-06 run and comparing it to the committed one — two independent runs of the same config — 0 of 7104 values differ by more than 0.1, 326 by more than 0.01, the largest being 0.076 on `PMSL.BIAS`/`MAE`/`RMSE`. That is the empirical basis for the `1e-1` tolerance: a stricter bound would make the test flaky without catching more, and it is also why the newer numbers were not committed.

`varda-single-1.0.yaml` **fails**, for reasons that predate this PR and are unrelated to it — the config and its reference come from the base branch. On 2026-08-03 it failed on 2 of 3326 metrics (`SP_10M`, max diff `1.10e-1` against a `1e-1` tolerance); on 2026-08-06 its experiment exited 1 after 20 minutes with a snakemake job failure, whose rule is not identifiable from the truncated CI stderr. It is in `CONFIGS`, so the heavytest job is red on varda while forecasters-ich1 passes.

**Regeneration end to end.** The regeneration path ran in CI on 2026-08-06 (`--regenerate-expected=all`, the flag form): forecasters-ich1 regenerated successfully, varda aborted because its experiment failed. Since the move to the script the generator has not been rerun end to end; it reproduces the committed `expected/forecasters-ich1.yaml` byte-for-byte from that run's `.nc` files, and the 18 unit tests cover its paths.

## Known limitations

- The second forecaster's checkpoint is a personal scratch path, `/scratch/mch/apennino/output/checkpoint/973b78b8.../inference-last.ckpt`. The test breaks if it disappears; it should move to mlstore like the first one.
- `expected/varda-single-1.0.yaml` is still the flat-list format, kept readable by the back-compat branch above.
- The generator runs the experiment itself rather than reading an existing output tree: `run_id` is a hash computed inside snakemake and the `.nc` files carry no config provenance, so which outputs belong to a config is only knowable from a pre-run mtime snapshot. A `--run-dir` escape hatch would fix that; not done here.


Louis-Frey:
cscs-ci run

Louis-Frey:
cscs-ci run heavytest

Louis-Frey:
cscs-ci run

Louis-Frey:
cscs-ci run heavytest

Louis-Frey:
cscs-ci run heavytest

workflow/rules/plot.smk
```diff
@@ -169,10 +169,26 @@ rule make_forecast_animation:
         """
 
 
+rule plot_prefetch_natural_earth:
```
lclanzi, line 172: What do you think about adding this change as a separate PR on evalml? Or is this change necessary for the success of the heavytest you designed?

Louis-Frey: Yes, I was thinking about that as well. I kept it here because it's almost strictly necessary for the forecaster heavytest to work and so far has not surfaced on main (most likely because nobody except myself (warm cache) yet has produced (many parallel) map plots). 

So it's good if it will be merged into main eventually, but I think there is no hurry.

tests/integration/test_configs.py
```diff
@@ -37,59 +43,91 @@ def load_expected(config_name):
...
 #     skip_suffixes = (".max", ".mean", ".min", ".std")
-#     metrics = [
-#         v for v in sorted(ds.data_vars)
-#         if "source" in ds[v].dims and not any(v.endswith(s) for s in skip_suffixes)
-#     ]
-#     regions   = ds.coords["region"].values.tolist()
-#     seasons   = ds.coords["season"].values.tolist()
-#     init_hrs  = ds.coords["init_hour"].values.tolist()
 #
 #     for config_name in CONFIGS:
```
lclanzi: This commented out part was left to indicate how to generated the expected config for varda-single. Does this commented out code now generate both expected configs (varda-single and forecaster)? If not, I would suggest adding the two code separately. Maybe comment them out in this script is not the best idea, we can see if we find better alternatives. What do you think?

workflow/rules/plot.smk
lclanzi: In theory, for the tests, there is no necessity for generating plots. Not sure whether there exist a flag which avoid the save-to-disk bit.

If you feel that the heavytest runs without this change, I would be in favor of adding this feature in a separate PR

Louis-Frey: First of all: No, there is no flag to turn off map plotting for scoremaps without turning off scoremaps functionality altogether. (Except by calling individual rules only, not `evalml experiment`.) 

In my understanding, the heavytest should test as much functionality of the repo as possible, that's why I enabled map (and other plotting like scorecards) in the config (and also fanned out along many of the configuration axes). If we think that testing plotting is out of scope for the heavytest, we can either (a) not test plotting at all, or (b) test plotting in a differently scoped test set / pipeline. So far the scope for (b) does not exist (except for the longtests, where plotting is also tested, but only very selectively), and I assumed we want to test pretty much everything in the heavytest, which is why I rejected (a). 

Also, the pipeline on both Varda + Forecaster combined did not even take that long (roughly 1h, max 2h from memory, I cannot access the log file any more). 

Why do you think that 
> In theory, for the tests, there is no necessity for generating plots.

? Do you worry about disk clutter? 

But I'm open for discussion, let me know.

tests/integration/test_configs.py
Louis-Frey: Good point. I'll look into it.

workflow/rules/plot.smk
lclanzi: I fully get your point. I was just wondering whether it is necessary to save the plots to disk once generated. But for the sake of testing the whole software, it also makes sense to do so, you are right.

Louis-Frey: Do the testing outputs get deleted eventually? If yes, I guess we can live with a couple of 100 .png's saved to disk. If not, maybe we can modify the cleanup script?

lclanzi: I assume the testing outputs get deleted, but something to check. I believe you could save the plots for now, and we will see if we need to change the behaviour in the future.

Maybe only try to double check whether it is better to keep this new rule here or opening a separate PR for it

Louis-Frey: I'll try to find out anyway if they are saved. If not (or after too long), you are right, our pipeline would produce quite a bit of clutter.

Louis-Frey:
cscs-ci run heavytest

Louis-Frey:
cscs-ci run heavytest

Louis-Frey:
cscs-ci run heavytest

tests/integration/test_configs.py
Louis-Frey: Good catch — no, it didn't generate both, and it couldn't have.

The `for config_name in CONFIGS` loop only picked the output *filename*: the `nc_files` glob inside it swept all of `output/data/runs/**` regardless of config, so both files would have been written the same content (every source found in the tree), and the `checked_sources == set(expected)` assert would then have failed for both.

You're also right that commenting it out wasn't the best idea, and the state of the block is the argument for it. Being dead code, two further defects had accumulated unnoticed: it called `math.isfinite()` without importing `math`, and it duplicated the test's own selection logic — the truth-source filter, the `.mean("step")`, the source-key derivation — with nothing keeping the copies in sync.

So rather than splitting it into two separate blocks, I made it live code with a single shared path, selected per config (c936ca7):

```
pytest tests/integration/test_configs.py -m heavytest --regenerate-expected=varda-single-1.0.yaml
```

Repeat the option for several configs, or pass `=all`. Comparison and regeneration now read values through the same `_metric_value()` helper, so an expected file cannot be generated with a different selection than the one asserted against — which is exactly what let the old block rot.

The flag takes an exact config name rather than relying on `-k` (6b6835c), because `-k` is substring-based: `-k varda-single` would also match a future `varda-single-2.0.yaml`, and `-k forecasters-ich1` already matches `forecasters-ich1-oper.yaml`. Three mistakes are rejected before the experiment starts: a bare `--regenerate-expected` (regenerating every config costs hours of GPU each, so it must be spelled `=all`), a partial name (with the valid names listed), and selecting no metric test at all (forgetting `-m heavytest` would otherwise report a green "N passed" having regenerated nothing). Only the `.nc` files a run actually rewrote are used, tracked by mtime snapshot, since `output/data/runs/` is shared across configs.

The third commit (104624a) sorts the `.nc` glob, so the source keys are written in a stable order — `glob` returns filesystem order, which would otherwise reorder the keys of a committed expected file whenever a multi-run config is regenerated on a different machine, producing a large diff carrying no value changes. It also re-writes `expected/forecasters-ich1.yaml` in that order, so the committed file is byte-for-byte what the flag produces; that commit's 6894/6894 diff is purely the two key blocks swapping places, with no value touched.

Verified on balfrin: `forecasters-ich1` regenerated with structure identical to the committed file — same source keys, in the same order, same sel rows, same metric names — and 0 of 7104 values outside tolerance (median drift 6e-05, max 0.076). That incidentally reproduces the tolerance rationale independently: 326 values drift past 0.01 and none past 0.1, closely matching the ~360 recorded in the comment.

Two things stay open. varda's expected file is still in the older flat-list format, so the `isinstance(expected, list)` back-compat branch has to stay for now. I couldn't regenerate it in that run because varda's experiment currently fails outright in `inference_execute` with `KeyError: 'FI_100'` in anemoi-inference's GRIB output (`write_initial_state`). It ran fine on 3 Aug, so this looks like environment drift rather than anything in this PR — probably worth its own issue. Once it's fixed, regenerating varda is a one-liner and the back-compat branch can go with it.

workflow/rules/plot.smk
Louis-Frey: Ok, according to the information by Claire on Teams, anything written to the CI temp dir in the service account should be wiped at the end of the CI run, so we don't have to worry about .png disk clutter.

tests/integration/test_configs.py
lclanzi: Thanks a lot, Louis. Just a few comments:

1. I like the idea of keeping the code used to generate the expected configs in functions rather than having it commented out. However, what do you think about moving all the functions that now generate the "ground truth" into a separate script instead of keeping them in `test_config.py`?

2. I see that the `plot_prefetch_natural_earth` rule is included in this commit. Have we decided to keep it here? I'm more inclined to put it in a separate PR, but either approach works for me.

3. The `KeyError: 'FI_100'` most likely arises from a new release of `anemoi-inference`, which was published two days ago. Pinning `anemoi-inference` to the previous release resolves this issue. In general, it may be a good idea to always pin `anemoi` and `earthkit` releases in the config or `pyproject` file and keep them up to date. However, I think this is out of scope for this PR.

Louis-Frey: Thanks for the suggestions Luca. 

1. Yes, we could do that. I will look into it.
2. Fair to question this point from a scoping perspective. We had some discussion about that further up, where I presented my reasons, and after your reply I assumed that you agree to this design. Let me know if you want to re-open this discussion. 
3. Yes, I am aware of that, but left it out on purpose because, as you say, it's not in the scope of this PR.

lclanzi: Thanks Louis :)

1. Thanks, if that will be possible, the end result may look cleaner.
2. Ok, I may have forgotten about it, sorry. Good to keep it here then
3. Yes, fully agree

Louis-Frey: On 1: Agreed — and done in `0b3f0a6`. The generator now lives in `tests/integration/regenerate_expected.py`:

```
python tests/integration/regenerate_expected.py forecasters-ich1.yaml
```

Name several configs to regenerate several, or pass `all`. `test_configs.py` only asserts now, so CI cannot rewrite a reference by any combination of flags.

Not a copy-paste, though — the old block rotted because it had its own copy of the selection logic. `tests/integration/expected_metrics.py` holds what both paths need (`find_nc_files()`, `run_experiment()`, `run_sources()`, `source_key()`, `metric_value()`) and both import it, so a reference still cannot be generated with a different selection than the one it is compared against.

Your suggestion also removed the whole `--regenerate-expected` machinery from `tests/conftest.py` — 108 lines rebuilding guard rails argparse gives for free, one of which covered a failure mode that only existed because the generator was a test.

`df6a457` adds unit tests, since moving the code out of `test_configs.py` also moved it out of everything that ran it. ~2 s against a synthetic dataset, mutation-checked rather than merely green. That exposed a real hole, fixed in the same commit: `setdefault()` creates a source's key before any row is built, so an all-NaN run produced `{source: []}`, which is truthy — the "refusing to write an empty reference" guard could not fire.

The generator still reproduces the committed `expected/forecasters-ich1.yaml` byte-for-byte, before and after that fix.

One limitation, unchanged: the script runs the experiment itself rather than reading an existing tree, since `run_id` is a hash computed inside snakemake and the `.nc` files carry no config provenance. A `--run-dir` escape hatch would fix that, but I haven't done it here.

lclanzi (approved):

PR #245, opened 2026-07-30
add stats when heavytest fail

lclanzi:
When heavytests fail, produce some stats. See below for example:

varda-single-1.0.yaml: 2898/3326 pass, 428/3326 fail (13%) | tol=1e-02 | diff range [0.00e+00 – 3.07e+00], median 4.69e-07 | failed vars: PMSL(72), SP_10M(140), TD_2M(48), TOT_PREC1(8), TOT_PREC6(14), T_2M(146)


clairemerker (approved):
Good idea! :)

PR #247, opened 2026-08-04
Disaggregation of temporally coarse forecasts results in NaNs

jonasbhend:
When disaggregating forecasts with 6hourly resolution to hourly, the necessary input data is not available. This PR fixes the default behaviour (error) to allow plotting of meteograms and dashboards for diverse datasets of forecasters (6-hourly) and temporal  downscalers (1-hourly).

## Summary of changes
* ensure only available lead times are loaded from grib files
* reindex before cumsum to avoid error

### DIFF
--- a/src/data_input/__init__.py
+++ b/src/data_input/__init__.py
@@ -1330,9 +1330,12 @@ def _disaggregated_and_derived_params(
                 cumuls[base] = _ensure_accum_ic(ds[base], load_steps)
             ds[agg_param] = _disaggregate_accum(cumuls[base], steps, n)
 
-    # Select only the originally requested steps (drop preceding helper steps)
+    # Select only the originally requested steps (drop preceding helper steps).
+    # Use reindex rather than sel so that steps absent from the loaded dataset
+    # (e.g. step 0 when the forecaster writes no step-0 file) are NaN-filled
+    # instead of raising a KeyError.
     if load_steps != list(steps) and "step" in ds.dims:
-        ds = ds.sel(step=[np.timedelta64(s, "h") for s in steps])
+        ds = ds.reindex(step=[np.timedelta64(s, "h") for s in steps])
 
     # Spatial derived params (skip if the loader already provided the variable natively)
     for p in params:
--- a/src/data_input/__init__.py
+++ b/src/data_input/__init__.py
@@ -458,10 +458,19 @@ def _collect_icon_archive_files(
             f"ICON-CH2-EPS): {root}"
         )
 
-    return [
+    all_paths = [
         reftime_dir / "grib" / f"{gribname}{lt // 24:02}{lt % 24:02}0000_{member_id}"
         for lt in steps
     ]
+    existing = [p for p in all_paths if p.exists()]
+    missing = [p for p in all_paths if not p.exists()]
+    if missing:
+        LOG.debug(
+            "Skipping %d archive file(s) not found (coarser source resolution?): %s",
+            len(missing),
+            missing,
+        )
+    return existing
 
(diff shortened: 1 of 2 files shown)


dnerini:
hi @jonasbhend, the code changes look good to me, but I'm not entirely sure I understand the problem. Is this issue affecting baselines or ML runs? can you provide a concrete example scenario that would trigger it?

src/data_input/__init__.py
```diff
@@ -458,10 +458,19 @@ def _collect_icon_archive_files(
...
 
-    return [
+    all_paths = [
         reftime_dir / "grib" / f"{gribname}{lt // 24:02}{lt % 24:02}0000_{member_id}"
         for lt in steps
     ]
+    existing = [p for p in all_paths if p.exists()]
+    missing = [p for p in all_paths if not p.exists()]
+    if missing:
+        LOG.debug(
```
dnerini, line 468: should there be some way to distinguish "expected coarser resolution" from "unexpected missing data"? should this be LOG.warning instead of debug to avoid that real data problems go unnoticed?

jonasbhend:
> hi @jonasbhend, the code changes look good to me, but I'm not entirely sure I understand the problem. Is this issue affecting baselines or ML runs? can you provide a concrete example scenario that would trigger it?

The concrete example is this config with a forecaster and a temporal downscaler being evaluated side-by-side (something we have done before). With explicit aggregations (i.e. TOT_PREC1 and TOT_PREC6), this will result in an error for both the experiment and meteogram use cases:

```
# yaml-language-server: $schema=../workflow/tools/config.schema.json
description: |
  Evaluate skill of Varda-single-1.0 against ground observations.

config_label: varda-single-1.0

dates:
  start: 2025-03-01T00:00
  end: 2025-03-03T00:00
  frequency: 24h

runs:
  - temporal_downscaler:
      checkpoint: https://service.meteoswiss.ch/mlstore#/models/sruc-m-2-interpolator/versions/3
      label: Varda-single-1.0
      steps: 0/120/1
      config: resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi.yaml
      extra_requirements:
        - anemoi-datasets==0.5.35
        # NOTE this pins the merge commit that introduced 'copy-prognostic-from-forecaster'
        # TODO pin a release tag once the plugins are versioned. Do not pin the main branch to avoid stale code (changes do to main to not change the evalml hash)
        - git+https://github.com/MeteoSwiss/anemoi-plugins-meteoswiss.git@b4835d76346e1e5ec0181884520da00e317fbfbc
        # - anemoi-inference==0.11.0
      forecaster:
        config: resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
        steps: 0/120/6
  - forecaster:
      label: Forecaster
      config: resources/inference/configs/sgm-multidataset-forecaster-global-ich1-oper.yaml
      steps: 0/120/6

experiment:
  params:
    - TOT_PREC1
    - TOT_PREC6
...
```

src/data_input/__init__.py
jonasbhend: @dnerini Looking into this. Are the other possible use cases for _collect_icon_archive_files than ICON archive, i.e. are we (planning to) use(ing) this for IFS or any other potentially non-hourly resolution forecast already? Otherwise, this change here could be dropped as ICON always has hourly grib files.

dnerini: no I suppose for now we're going to stick to icon archive data, hence hourly by design

jonasbhend: ok, revert now

PR #248, opened 2026-08-04
Feat/add tool compare inference

lclanzi:
Adds a Marimo interactive notebook to compare raw GRIB outputs from two `evalml` inference runs field-by-field.

The tool reads GRIB messages using `eccodes`, compares corresponding fields between two runs, and presents a `pandas` result table showing:

- Maximum absolute difference
- Mean absolute difference
- Pass/fail status based on configurable tolerance thresholds

## Usage workflow

1. Run `evalml` and generate an inference output.
2. Rename the output folder to preserve the first run.
3. Run `evalml` again and generate a new inference output.
4. Launch the comparison notebook:

```bash
uv run marimo run tools/compare_grib.mo.py
```
5. Provide path of two evalml runs. Examples:
  PathA = /scratch/mch/llanzila/test/evalml/output_ground_truth/data/runs/temporal_downscaler-sruc-a157-on-forecaster-sruc-94fd/21c5/202501010000/grib
  PathB = /scratch/mch/llanzila/test/evalml/output/data/runs/temporal_downscaler-sruc-a157-on-forecaster-sruc-94fd/21c5/202501010000/grib

An example of the output is shown below

Screenshot 2026-08-04 110023


jonasbhend (commented):
Cool addition and very nice summaries of the differences.

Would it make sense to isolate the logic for equivalence testing of grib files into a module for use in testing?

lclanzi:
I think the challenging part is that this type of testing always requires two evalml outputs, one serving as the reference (ground truth/baseline) and one generated with the current version. I'm not sure how maintainable that would be in the long term, and it is also not straight forward to implement imo.

That's why I added it as a standalone tool rather than integrating it into the test suite, but I'm happy to discuss other approaches

PR #249, opened 2026-08-07
Prefer current station location in StationCatalog.from_meta

clairemerker:
DWH station metadata has one row per (parameter, operational period), each with its own coordinates, e.g. here for station GEN:

```
parameter  variable    from        to          latitude   longitude  elev
dkl010z0   DD_10M      1892-07-24  1939-08-31   45.931282  9.019816   1701
dkl010z0   DD_10M      2012-10-23  current      45.927593  9.017875   1600
prestas0   PS          1992-03-26  2012-10-23   45.926235  9.014575   1608
prestas0   PS          2012-10-23  current      45.927593  9.017875   1600
rre006i0   TOT_PREC6   1892-07-24  1939-08-31   45.931282  9.019816   1701
rre006i0   TOT_PREC6   2016-03-03  current      45.927593  9.017875   1600
tde200s0   TD_2M       1994-07-14  2012-10-23   45.926365  9.014614   1608
tde200s0   TD_2M       2012-10-23  current      45.927593  9.017875   1600
tde200s0   TD_2M       2026-12-31  2026-12-31   45.927593  9.017875   1600
``` 

`from_meta()` collapsed these by sorting alphabetically and keeping the parameter's oldest op_since row, which could lead to non-deterministic metadata.

This PR select instead by (current op_till first, fixed parameter priority, latest op_since), falling back to the latest period when no current row exists. This yields the station's current coordinate, is deterministic and stable against DWH row churn, and still returns one coordinate per station.


lclanzi (approved):
Thanks a lot Claire, looks good to me!

You could maybe just add in the description of the from_meta() function that currently the parameter priority is randomly selected, and this could be improved in the future (e.g. each param carries its own metadata). What do you think?

PR #252, opened 2026-08-12
Scale the number of physical cores requested with the expected memory usage

dnerini:
On balfrin there is no memory control, in principle every job can use the whole memory of the node, e.g. 512 GB - which may lead to job failures in case the total memory is exceeded.
As workaround, it was proposed to user who have jobs using large amount of memory, e.g. more than 2 GB, to scale the number of physical cores requested, with the expected memory usage (even if you are using only one effectively). 

This PR applies this guideline to evalml where relevant.

FYI @lxavier

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -469,6 +469,22 @@ PARAMS_WITHOUT_STEP_ZERO_VALUE = {
 }
 
 
+def balfrin_cpus(mem_mb, actual_cpus=1, node_mem_mb=512_000, node_cpus=256):
+    """Minimum cores to request on Balfrin based on memory usage.
+
+    Balfrin CPU nodes have no per-job memory control: any job can consume the
+    full node memory (512 GB / 256 cores).  To avoid starving other jobs,
+    reserve cores proportional to expected memory so the scheduler implicitly
+    limits total memory usage:
+
+        cores = max(actual_cpus, ceil(mem_mb / node_mem_mb * node_cpus))
+    """
+    import math
+
+    mem_cpus = math.ceil(mem_mb / node_mem_mb * node_cpus)
+    return max(actual_cpus, mem_cpus)
+
+
 def resolve_leadtimes(steps_spec, requested="all", param=None):
     """Lead times to compute for a single participant.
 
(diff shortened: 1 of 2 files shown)


Louis-Frey (approved):
Hi Daniele, thanks for taking care of this!

While I think the PR is rather self-explanatory, I still gave it to Claude for an adversarial review, which surfaced a couple of aspects. After some discussion with Claude, I discarded some of the aspects, but two remain:

## 1. The extra cores are not idle — they raise actual memory usage

`verify()` sizes its dask thread pool from the allocation
(`num_workers = len(os.sched_getaffinity(0))`, `src/verification/__init__.py:363`), and
`verification_metrics.py:93` calls it without `num_workers`. With `ConstrainCores=yes` in cgroup.conf the
affinity mask equals the allocation, so `verification_metrics` and `verification_metrics_baseline` go from
24 to **40 concurrent dask threads**. Peak RSS in the threaded scheduler grows roughly with the number of
in-flight chunks, so real consumption rises by up to ~1.7× while `mem_mb=80_000` stays put — i.e. the
reservation is sized on a memory figure the change itself invalidates.

The guideline's "even if you are using only one effectively" does not hold for this code, and nothing
enforces `actual_cpus`. Cheapest fix: pass the intended count through, e.g.
`--num_workers {params.actual_cpus}` in the two shell commands. Otherwise `mem_mb` needs re-measuring at
40 threads.

(`verification_scoremaps` is eager numpy/xarray — its 2 → 25 bump changes the reservation only.)


## 2. `run_mec` and `run_ffv2`

These rules are unaddressed and I am not familiar with their memory behaviour. Maybe no problem at all, I thought I'm gonna raise it anyway. 


Otherwise good to merge from my side.

dnerini:
hi @Louis-Frey, thanks. I think the memory requirements for the verification rules where quite generous to start with, so I don't expect this to be a problem. But it's a good point and I should at least check it before merging. Will run some tests when I find some time!

Louis-Frey:
Another issue was that effective memory available on the nodes is not 512'000 MB but lower, but maybe we don't want to be that picky.

PR #257, opened 2026-09-04
Change order of operations for ensemble mean of derived parameters

jonasbhend:
For all derived quantities, the order of operation to compute the ensemble mean needs fixing. First we need to compute the derived parameter per ensemble member, then average (now we're doing it in reverse).

For aggregated parameters, this is a non-issue as long as we only compute the ensemble mean (not median or other ensemble statistics).

## Summary of changes
* change order of operations for derived parameters


cosunae (approved):
good catch. 
I checked that the derived params wont be computed later on, if variables are already in ds.data_vars so this looks good to me

PR #258, opened 2026-09-04
Paper figures improvements

clairemerker:


config/varda-single_meteogram_20250321.yaml
jonasbhend, line 12:
```suggestion
  - 2025-03-21T00:00
```
I think this emphasizes that a single initialization is run.

config/varda-single_meteogram_20250627.yaml
jonasbhend, line 11:
```suggestion
  - 2025-06-27T18:00
```
same as above

docs/publication_figures.md
jonasbhend, line 251:
```suggestion
- *Needs:* truth dataset manifest(s) with scoremap NC files (produced
```
scoremaps.py also supports adding scores per station on top of map, or standalone map with just stations.

workflow/scripts/report_scorecard.py
jonasbhend, line 1: I am a bit worried that changes to the 'original' scorecard script affect the look of this? Have you checked if the standard scorecard still looks the same / acceptable?

jonasbhend (approved):
Hi @clairemerker. Thanks a lot for the many improvements. I have only minor comments and am happy to merge once these are addressed.

PR #259, opened 2026-09-07
Add station-group stratification 

lclanzi:
This PR adds a new stratification axis, station group (`all` / `holdout` / `holdin`), alongside the existing region/season/init_hour stratifications, plus supporting infrastructure to make ground-truth retrieval more flexible. Note that this is a single fixed holdout/holdin split used for stratified evaluation, not k-fold cross-validation.

- verification metrics can now be split by `all` / `holdout` / `holdin` stations, so skill can be evaluated separately on stations that were withheld from nudging (or other applications) vs. those that fed into it.
- `experiment.cross_validation` in the top-level config (`exclude_stations`, or `holdout_fraction` + `holdout_seed` for a random split) is the single source of truth for which stations define the holdout group
- `verify()` (`src/verification/__init__.py`) adds a `station_group` dimension to verification outputs, alongside the existing region/season/init_hour dimensions.
- The dashboard (`report_experiment_dashboard.py`, `template.html.jinja2`, `script.js`) and scorecards (`report_scorecard.py`) are updated to expose and correctly aggregate this new dimension.
- `verif_hash()` in `common.smk` now includes the station-group configuration, so changing it correctly invalidates cached verification outputs instead of silently reusing stale results.
- Geographic station filtering for jretrieve-based truth: `data_input.jretrieve.parse_selection()` now supports `filter_mode=switzerland` (trim to stations within the real Swiss national border) and `filter_mode=domain;domain_bbox=...` (trim to an arbitrary bounding box), plus an optional `use_limitation=<N>` selector for data-usage-rights filtering. This lets a truth-root retrieve over a broad domain and then restrict to a precise region before verification


src/data_input/__init__.py
```diff
@@ -758,7 +814,28 @@ def load_obs_data_from_jretrieve(
...
+        "TOT_PREC": "tp",
+        "VMAX_10M": "vmax",
+    }
+    n_total = result.sizes["values"]
+    for icon, short in _icon_to_short.items():
+        if icon in result.data_vars:
+            n_valid = int(result[icon].notnull().any("time").sum())
+            LOG.info(
+                "Stations with valid %s: %d / %d stations", short, n_valid, n_total
+            )
```
jonasbhend: I would have expected to see COSMO names here (instead of ifs ones), so no need for translation, in particular since we are now going DWH names --> COSMO names --> IFS names, which is a bit much.

jonasbhend: And this is 'just' for the log message

jonasbhend (commented):
Hi @lclanzi. Thanks for implementing this. I think the 'cross-validation' with holdin/out stations is a very nice feature. Given the additional way to select stations from jretrieve, I was wondering if we really want to support bbox selection and filtering down to Switzerland? What is the use case here that is not covered by selection by measurement network (e.g. 1,2) or bbox? If we don't have a strong argument, I would rather avoid this additional complexity.

Also, I am unable to check the implementation on linard, as mlflow is still being blocked. Will try to do a short test on Wednesday to confirm everything is working as expected.

lclanzi:
> Hi @lclanzi. Thanks for implementing this. I think the 'cross-validation' with holdin/out stations is a very nice feature. Given the additional way to select stations from jretrieve, I was wondering if we really want to support bbox selection and filtering down to Switzerland? What is the use case here that is not covered by selection by measurement network (e.g. 1,2) or bbox? If we don't have a strong argument, I would rather avoid this additional complexity.
> 
> Also, I am unable to check the implementation on linard, as mlflow is still being blocked. Will try to do a short test on Wednesday to confirm everything is working as expected.

Hi @jonasbhend, thanks a lot for the feedback! 
I see your point, and it makes sense. However, I wasn’t able to retrieve SwissMetNet + partner stations (limited to Switzerland) directly through the jretrieve interface. Therefore, I currently retrieve all stations within the bounding box and filter them afterwards to keep only those in Switzerland. If you know how to achieve this directly with jretrieve, I will remove this additional complexity.

jonasbhend:
I know that this doesn't work (easily), but my question is: why do you want to use only stations in Switzerland? I guess for the nudging, you wouldn't want to constrain, and for the evaluation we could argue similarly. If anything, I think for the evaluation we should implement it as a stratifiable region (in essence the hull of all forecasting regions), which would then be available for both station and analysis-based evaluations.

src/data_input/__init__.py
lclanzi: I removed this conversion, it was AI slop. Thanks

lclanzi:
Thanks for the suggestions @jonasbhend!
I have added a new shapefile for Switzerland. This region is now available in the list and it is used for applying a region stratification over the country (instead of the filter in jretrieve)

workflow/rules/inference.smk
```diff
@@ -312,20 +318,22 @@ rule inference_execute:
...
-                    --ntasks={resources.ntasks} \
-                    anemoi-inference run config.yaml "${{CMD_ARGS[@]}}"
+                anemoi-inference run config.yaml "${{CMD_ARGS[@]}}"
             }}
             export -f _run_inference
 
-            squashfs-mount {params.env_path}:/user-environment -- bash -c '_run_inference /user-environment'
+            srun \
+                --job-name=anemoi-inference \
+                --uenv={params.env_path}:/user-environment \
```
dnerini, line 327: the PR includes the changes needed for Linard, correct? so it should be merged only after #263 ?

resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi_with_nudging.yaml
```diff
@@ -0,0 +1,189 @@
...
+                  use_limitation: 40
+                  run_mode: "devt"
+                  station_filter_mode: domain
+                  trim_bbox: [45.7, 48.0, 5.8, 10.8]
+            - forward_transform_filter:
+                clean_observation:
+                  obs_path_in: observation/nudging_station_obs_raw.parquet
+                  obs_path_out: observation/nudging_station_obs.parquet
+            - forward_transform_filter:
+                nudge_toward_observation:
```
dnerini, line 32: this PR depends on https://github.com/MeteoSwiss/anemoi-plugins-meteoswiss/pull/40 right?

resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi_with_nudging.yaml
```diff
@@ -0,0 +1,189 @@
...
+                  obs_path: observation/nudging_station_obs.parquet
+                  nudge_variables: [T_2M, TD_2M, U_10M, V_10M, PMSL, PS]
+                  temperature_lapse_rate: 0.0065  # K/m
+                  pressure_lapse_rate: 11.5  # Pa/m
```
dnerini, line 36: do we need to mention all parameters here? for sake of simplicity, could we exclude all those using defaults?

resources/report/dashboard/script.js
```diff
@@ -327,6 +332,7 @@ function resizeChartScroll() {
       getSelected("source-select"),
       getSelected("metric-select"),
       getSelected("param-select"),
+      stationGroupSelect ? [stationGroupSelect.value] : [],
```
dnerini, line 335: why is this entry formulated differently compared to the others?

src/data_input/jretrieve.py
```diff
@@ -299,7 +328,7 @@ def fetch_data(
     increment_minutes=60,
     seq_type="surface",
     stage="prod",
-    use_limitation: int = 40,
+    use_limitation: int | None = None,
```
dnerini, line 331: here the default value for `use_limitation` changed to `None`, isn't this going to affect pre-existing experiment configs?

src/verification/__init__.py
```diff
@@ -386,6 +403,17 @@ def verify(
         lon=obs_aligned["longitude"], lat=obs_aligned["latitude"]
     )
 
+    station_masks = None
+    if holdout_stations is not None:
```
dnerini, line 407: here an empty list will count as a populated one and trigger the all/holdout/holdin stratification

src/evalml/config.py
```diff
@@ -455,6 +455,29 @@ class Dashboard(BaseModel):
...
 
 
+class StationHoldoutConfig(BaseModel):
+    """Station holdout settings for station-group stratified verification."""
+
+    holdout_fraction: Optional[float] = Field(
+        default=None,
+        description=(
+            "Fraction of truth stations to hold out for evaluation (exclusive 0–1). "
+            "Mutually exclusive with exclude_stations."
```
dnerini, line 465: how is this enforced? it looks like the code later on prioritizes `exclude_stations` over  `holdout_fraction` without any warnings or documentation

config/varda-rapid-0.2.yaml
```diff
@@ -0,0 +1,150 @@
...
+    - T_2M
+    - TD_2M
+    - U_10M
+    - V_10M
+    - TOT_PREC
+    - PMSL
+    - PS
+  stratification:
+    regions:
+      - switzerland
```
dnerini, line 71: this is not supported in https://github.com/MeteoSwiss/evalml/blob/24b5b02cd1d71efecbdd49e618a01e3361553f87/workflow/rules/common.smk#L82, where is it coming from?

.gitignore
```diff
@@ -14,7 +14,7 @@ _dev
 .vscode
 .idea
 .snakemake
-output
+output*/
```
dnerini, line 17: anyone using `output` as a symlink now risks `git add -A` staging the symlink

workflow/scripts/inference_prepare.py
```diff
@@ -170,6 +191,50 @@ def _override_recursive(original: dict, updates: dict) -> dict:
     return original
 
 
+def _inject_nudging_station_holdout(config: dict, station_holdout_cfg: dict) -> None:
```
dnerini, line 194: I would try to keep the station holdout logic independent from the usage done downstream: this feature might be useful for other use cases other than nudging!

workflow/scripts/report_experiment_dashboard.py
```diff
@@ -89,8 +96,9 @@ def main(args):
         df = df[df["season"] == "all"]
     if "init_hour" not in stratification:
         df = df[df["init_hour"] == "all"]
+    if "station_group" not in df.columns:
```
dnerini, line 99: this check needs to be included in all scripts consuming data from verify().

a safer and more convenient approach would be to have verify() in verification_metrics.py to always emit station_group=["all"] uniformly

src/evalml/config.py
```diff
@@ -455,6 +455,29 @@ class Dashboard(BaseModel):
     )
 
 
+class StationHoldoutConfig(BaseModel):
+    """Station holdout settings for station-group stratified verification."""
+
+    holdout_fraction: Optional[float] = Field(
```
dnerini, line 461: adding a pydantic field constraint would help validate the value for the range [0,1]

workflow/scripts/verification_metrics.py
```diff
@@ -31,6 +32,27 @@ class ScriptConfig(Namespace):
...
+    holdout_fraction = station_holdout_cfg.get("holdout_fraction")
+    holdout_seed = station_holdout_cfg.get("holdout_seed", 42)
+
+    if exclude_stations is not None:
+        return [s for s in exclude_stations if s in all_stations]
+    if holdout_fraction is not None and 0.0 < float(holdout_fraction) < 1.0:
+        n_holdout = round(len(all_stations) * float(holdout_fraction))
+        rng = np.random.default_rng(holdout_seed)
+        return list(rng.choice(all_stations, size=n_holdout, replace=False))
+    return []
```
dnerini, line 53: why should an invalid fraction silently return an empty list here?

dnerini (commented):
very cool feature! I left some comments

workflow/rules/inference.smk
lclanzi: Correct. Although this change also works on Balfrin, removing the dependency from https://github.com/MeteoSwiss/evalml/pull/263

resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi_with_nudging.yaml
lclanzi: Yes, correct. We should merge it afterwards

resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi_with_nudging.yaml
lclanzi: My idea was to limit the number of hard-coded params. However, if you think that `temperature_lapse_rate` and `pressure_lapse_rate` should be treated as constants, I will then remove them from the config

resources/report/dashboard/script.js
lclanzi: AI slop. The entry is now formulated like the other ones. This only had a minor impact on the dashboard formatting

src/data_input/jretrieve.py
lclanzi: I changed the `use_limitation` default back to 40

src/verification/__init__.py
lclanzi: Thanks for spotting this. I now set it to `if holdout_stations:`, so that an empy list does not trigger the station stratification

src/evalml/config.py
lclanzi: It wasn't enforced, thank for spotting it. Two separate places (compute_holdout_stations() and _inject_nudging_station_holdout()) silently preferred exclude_stations over holdout_fraction if both were set, with no warning. I've now added a Pydantic validator on StationHoldoutConfig that raises an error if both are set, so it's enforced at config-load time.

config/varda-rapid-0.2.yaml
lclanzi: I had to perform a couple of tests, and therefore added the shapefile and support for this region. Note that this did not require any code change, but solely uploading the shapefile to the correct path

.gitignore
lclanzi: Correct. I removed the trailing slash

workflow/scripts/report_experiment_dashboard.py
lclanzi: Agreed and fixed. `verify()` now always emits a` station_group` dimension (["all"] when no holdout is configured, ["all", "holdout", "holdin"] when it is), so downstream scripts no longer need to check for its presence

src/evalml/config.py
lclanzi: Makes sense. I added gt=0, lt=1 to holdout_fraction's Pydantic Field, matching the range already documented in its description. Out-of-range values (0.0, 1.0, 1.5, negative) now fail at config-load time

workflow/scripts/verification_metrics.py
lclanzi: I added a check with Pydantic, so that an out-of-range value trigger immediately an error, therefore avoiding that an invalid fraction return an empty list

config/varda-rapid-0.2.yaml
dnerini: oh I see, thanks, very well. Could you just make sure that the shape files have the same permissions as the other ones in that same folder?

resources/inference/configs/sgm-temporal-downscaler-global_trimedge_multi_with_nudging.yaml
dnerini: yes, I think it'd be good to simplify somewhat the interface by removing some of the parameters that are less likely to be used

lclanzi: Agree. I removed several params, and set them as default

workflow/scripts/inference_prepare.py
lclanzi: Good idea. I disentagled the nudging from the station stratification. The function `_inject_nudging_station_holdout` now gets executed only when the nuding filter is in the inference config

src/data_input/jretrieve.py
```diff
@@ -168,20 +168,40 @@ def _stations_to_argv(stations: dict[str, Any]) -> list[str]:
...
       ``jretrievedwh:locations=ARO,KLO``
       ``jretrievedwh:bbox=45.8,47.8,5.9,10.5``
+      ``jretrievedwh:bbox=40.5,53.0,0.0,17.5;use_limitation=40;filter_mode=switzerland``
+        -> retrieve over the given bbox, then trim to stations within the real
+           Swiss national border (see load_obs_data_from_jretrieve).
+      ``jretrievedwh:bbox=40.5,53.0,0.0,17.5;filter_mode=domain;domain_bbox=45.7,48.0,5.8,10.8``
+        -> retrieve over one bbox, then trim to a second, independent bbox —
+           mirrors RetrieveObservation's bbox/station_filter_mode+domain_bbox split.
+
+    use_limitation defaults to 40 when not given, matching fetch_data()'s own
```
dnerini: I suggest to document the current behavior only, removing any reference to previous versions and its changes

PR #260, opened 2026-09-09
Mount venv squashfs via srun --uenv instead of squashfs-mount

dnerini:
On Linard, the cluster now provides a native `srun --uenv=<squashfs>:<mountpoint>` flag (via the `slurm-uenv-mount` SPANK plugin), which mounts the squashfs inside the task's own namespace on the allocated node.

### DIFF
--- a/workflow/rules/inference.smk
+++ b/workflow/rules/inference.smk
@@ -293,7 +293,7 @@ rule inference_execute:
             cd {params.workdir}
 
             _run_inference() {{
-                local VENV=$1
+                local VENV=/user-environment
                 source "$VENV/bin/activate"
 
                 if [ "{params.disable_local_definitions}" = "False" ]; then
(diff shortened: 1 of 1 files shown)


lclanzi:
Works nicely, thanks!

Just a small comment. The job currently shows up as "bash" in the queue. Adding " --job-name=anemoi-inference \ " to the _srun_ command would override the default naming and make it easier to identify.

dnerini:
> Just a small comment. The job currently shows up as "bash" in the queue. Adding " --job-name=anemoi-inference \ " to the _srun_ command would override the default naming and make it easier to identify.

Nice! Can you suggest it directly as a change in the code?

workflow/rules/inference.smk
```diff
@@ -307,19 +307,20 @@ rule inference_execute:
...
-                    --ntasks={resources.ntasks} \
-                    anemoi-inference run config.yaml "${{CMD_ARGS[@]}}"
+                anemoi-inference run config.yaml "${{CMD_ARGS[@]}}"
             }}
             export -f _run_inference
 
-            squashfs-mount {params.env_path}:/user-environment -- bash -c '_run_inference /user-environment'
+            srun \
+                --uenv={params.env_path}:/user-environment \
+                --unbuffered \
```
lclanzi, line 317:
```suggestion
            srun \
                --job-name=anemoi-inference \
                --uenv={params.env_path}:/user-environment \
                --unbuffered \
```

lclanzi (commented):
Include suggestion

lclanzi (approved):

PR #261, opened 2026-09-09
Add check for missing steps in _disaggregate_accum

dnerini:
By accident, I requested results for TOT_PREC1 while only running a 6-h forecaster. This triggered an obscure error message from deep down in pandas. Instead, this PR introduces a simple check upfront to provide a clearer error message for the same scenario.


jonasbhend:
I propose to instead implement #247, that way we can evaluate across diverse forecasts without triggering an error.

PR #180, opened 2026-06-09
Expose SLURM account and GPUs from config

jonasbhend:
The slurm account is not passed on to snakemake options. 

### Summary of changes
* add slurm_account and gpus as optional fields in config schema
* rename to `slurm_account` in temporal downscaler config

PR #262, opened 2026-09-09
fix: pass slurm account to inference srun call

lclanzi:
## Summary
- The `inference_execute` rule runs `anemoi-inference` via a nested `srun` inside a `squashfs-mount ... bash -c` block, separate from snakemake's own job submission
- `slurm_account` (added in #180) was only wired into snakemake's `--default-resources`, so it never reached this inner `srun` call. This meant that GPU inference jobs ran under the cluster's default account instead of the one configured in `profile.default_resources.slurm_account`
- Adds an `account_flag` param that renders `--account=<slurm_account>` when configured, or an empty string otherwise, and passes it to the inner `srun` call.


frazane (approved):
LGTM, thanks!

PR #263, opened 2026-09-10
Migrate evalml to Linard (new Balfrin)

dnerini:


lclanzi:
FYI: mounting venv squashfs via srun also works on Balfrin

dnerini:
> FYI: mounting venv squashfs via srun also works on Balfrin

good to know! it means we can already merge these changes then?

lclanzi:
maybe double check, but otherwise yes. It works on my end

PR #265, opened 2026-09-18
Add verification support for surface variable added with post-processors

lclanzi:
Adds wind speed (SP_10M), wind direction (DD_10M) and relative humidity (RELHUM_2M) as verifiable parameters,

- Reads `SP_10M`/`DD_10M`/`RELHUM_2M` natively from ML-inference GRIB. Derives them from `U_10M`/`V_10M`/`T_2M`/`TD_2M` for baselines (ICON archive, INCA).
- Fetches `RELHUM_2M` natively from the DWH (`ure200h0`) instead of deriving it from `T_2M`/`TD_2M`
- Adds circular-aware `BIAS`/`MSE`/`MAE`/`CORR` for `DD_10M`: a plain difference near due-north (e.g. 355° vs 5°) previously measured a 350° error instead of the true 10°.
- Adds a `RELHUM_2M` colormap default (viridis, 0-100%)


jonasbhend (commented):
Hi @lclanzi. Very cool feature indeed. 

I suggest we discuss if we want to ditch derived parameter computation in evalml in favor of using post-processors (and ignore sources for which derived parameters are unavailable). What do you think?

My second comment relates to direction of wind. For this, the standard verification metrics do not make sense (e.g. mean absolute error) and should be replaced with error metrics for circular quantities. I suggest to at least include a warning, or maybe ditch DD_10M for now and comment out the relevant sections.

lclanzi:
Thanks for the feedback @jonasbhend!

Yes, the idea is to compute all variables during inference, in order to minimize external operations.

For DD_10M:

- Error metrics (bias, RMSE, MAE, correlation): before computing the error, I shift the observation to whichever "lap" around the circle is closest to the forecast. So forecast 355° vs. observed 5° reads as a 10° error, not 350°.
- Mean/variance: instead of averaging degrees directly, I average the direction as a vector (its sin/cos components) and take the angle of that. mean(359°, 1°) comes out to ~0°.
- Min/max: I removed these for wind direction

